PR #81, opened 2020-06-01
minified race avatars

filipslanchev:


public/env.js
```diff
@@ -1,4 +1,4 @@
 window._env_ = {
-  BASE_URL: "https://statistic-service-test.w3champions.com/",
```
abeninskibede, line 2: Revert this

filipslanchev: done

filipslanchev:
finalized

PR #116, opened 2020-06-19
[Proposal] Optimize v-card & container

lennartblom:
Hi everyone! 👋 
Thanks for providing such an awesome community project to enjoy Warcraft III even more! You're the best! 🏅 

As a small idea, I would like to propose two minor changes to the look and feel of the overall page. 👀 

**Edit border radius**  🔘 

To polish the look and feel of the content area I propose adding a `border-radius` of `15px`, which is often used in UI elements. The hard corners feel a little bit forced and my perspective is the UI is feeling more appropriate with rounding it up a little bit. 

**Enable more space with removing padding for mobile screens** 📱 

Opening contents and sites of the w3champions pages on mobile devices is quite uncomfortable, because the site is providing much data to the viewer... one first step to mitigate problematic views could be removing the padding and gain `26px` of space overall 👾 

I summed up my changes with a set of screenshots. 

**Left:** current state of the UI. **Right:** My modified state of the UI.

modified

What do you think? Love to hear your feedback! 🚀 

Cheers from Hamburg!


modmoto:
Hi Lennart,

thanks for the PR, really nice to see our first PR like that =)

To be fair, I am not a big fan of the round edges, as we do not have them anywhere else. I dont think that this fits the overall picture.
Regarding the mobile UI: I also like the colored border that the background provides. Yes, it is not needed, but I do not see 16 Pixels make that much difference. But the Mobile UI was mostly in @abeninski s hands, so I only have an opinion there ;)

@abeninski what is your thought on that?

@lennartblom Don´t take this the wrong way, I am really glad for your proposals and would be happy if you have any other ideas, as none of us are real UI designers and every help or fresh ideas that we can get helps us =) We just practice a very honest feedback culture around our UI choices and this worked quite well so far.

modmoto

lennartblom:
Fair enough! 🙂 Thanks for the honest and direct feedback! I expected something like this but wanted to speak out my taste anyway, so I could hear your perspective regarding the padding and edges.

I think I might have some more ideas which are less based on taste. So perhaps I can be more successful 😈 

No surprise I dedicated my professional area in backend engineering... I did some years of freelance webdesign but the discussions about what's more elegant were always hard for me... in this kind of community OSS projects it's interesting though, because there's no monetary pressure around it and nice to communicate ideas with others! 

Keep up the good spirit 🙂 And thanks again!

PR #167, opened 2020-09-15
fixed flag z-level

Cepheid-UK:
Flag was sitting over link, preventing click, see pic
flag


modmoto:
@Cepheid-UK thanks by the way =D

PR #169, opened 2020-09-16
Slight improvement to FFA grid view

Cepheid-UK:
Modified some cell widths to fit more cases of Icon clipping on the FFA table, the pictures show

Before (Live):
before

After (PR):
after

Still not perfect, but better in my opinion.


Cepheid-UK:
Actually, I'd forgotten to take into account the other game modes, I'll revise this.

modmoto:
kk ^^ but if you have that one open, it would be good to just fix it for good, imo.

PR #187, opened 2020-10-08
fixed match page to show 4v4 heroes + scores properly

Cepheid-UK:
Fixed the match page to show all detail of a finished 4v4 match (was previously only showing top 2 players heroes and top 4 players score data)

heroes

scores

The scores still looks quite poor, and could still be improved later to look more like the FFA table.

f72cd75: Removed some out-of-date text on the Winrates page.


src/views/MatchDetail.vue
```diff
@@ -249,6 +235,13 @@ export default class MatchDetailView extends Vue {
     );
   }
 
+  get matchIs4v4() {
```
abeninskibede, line 238: Doesn't seem like this is used anymore. You can delete it

Cepheid-UK: f72cd75 done 👍

Cepheid-UK: 09eccd4 done 👍

src/views/MatchDetail.vue
```diff
@@ -233,15 +219,6 @@ export default class MatchDetailView extends Vue {
     return this.$store.direct.state.matches.matchDetail.match.season ?? 1;
   }
 
-  get matchIs2v2() {
```
Cepheid-UK, line 236: Also removed this as it's no longer used.

PR #188, opened 2020-10-09
fix vuetify deprecation warning

SergeyKhval:
Fix v-content deprecation warning
Add pre-commit hooks for lint and prettier


abeninski:
Hey!
Thanks for the PR.

I'm not really a fan of commit hooks as they tend to break the build after you checked it?
I think we should relax the rules for linting.

SergeyKhval:
Sure, I'll remove this part from PR

PR #195, opened 2020-10-15
Removed Home button + removed 2v2 AT from statistics page

Cepheid-UK:
- Home button removed from top bar (The w3champions logo acts as home)
- removed 2v2 AT from the list of game modes that can be queried in the Player Activity

ATremoved


modmoto:
why removing the at stas? They are different from the rt stats, or did that chang?

Cepheid-UK:
Since AT and RT has been merged the 2v2 AT stats have been recorded differently, if you check how this looks on the live site you'll see what I mean. All 2v2 stats are under '2v2' now. The stats service may also benefit from an update but I suck with c# haha.

modmoto:
ok, so the stat service should still get at-rt stats merged, right? Otherwise we show incorrect data (although it is no big difference, as AT is not played that much)

Cepheid-UK:
As I understand how @abeninski implemented it, what used to be GM_2ON2_AT is now merged with the stats in GM_2ON2
Which is why '2 vs 2 AT' shows 0 games played since the merge:

2v2AT-1

AT is played much more now since the merge! but many of those games are RT vs AT.

As far as I know, there's no flag to tell now if a game is RT vs RT, AT vs RT or AT vs AT. it's all just GM_2ON2.

We could keep the GM_2ON2_AT on the website for historic data if you think that is valuable, but in 2 months it will be past the x-axis limit and effectively that will show no stats.

PR #217, opened 2020-10-23
corrected to prod BASE_URL

Cepheid-UK:
Point back to prod.


Cepheid-UK:
Ignore this, works fine on deploy.

PR #220, opened 2020-10-25
added new css style for scrollable dialog and applied it in user icon…

jensk1:
I added a new css class for the dialog, which was not able to be scrollable by default somehow.

In the PlayerAvatar i let the dialog inherit from the scss. 

Please let me know what you think or if something needs to be changed.

The Result is that the you can scroll down when the height of the page gets too small.


Cepheid-UK:
Tested and working!

Thanks for the PR.

Issue #206, opened 2020-10-22
Users with non-race icons get "races.undefined" on hover


Cepheid-UK:
When you hover over a player's icon in the rankings, typically it will show "Undead" or "Orc" etc.
For players with special icons (rewards/devs) this shows "races.undefined"

Players currently with non-standard icons for testing:

https://www.w3champions.com/player/Cepheid%231467
https://www.w3champions.com/player/Kenshin%232209
https://www.w3champions.com/player/blaST9Z%232118

Chappitono:
The API for the player returns : 
{
	"calculatedRace": 0,
	**"selectedRace": 32,**
	"pictureId": 3,
},

in RankingsGrid.vue : 

private hasSelectedIcon(playerInfo: PlayerInfo) {
    if (playerInfo.selectedRace !== undefined && playerInfo.selectedRace != null && playerInfo.pictureId !== undefined && playerInfo.pictureId != null){
      **return playerInfo.selectedRace !== EAvatarCategory.TOTAL;**
    }
    return false;
  }

The method returns true since total is equal to 16. (I find it a little bit strange to compare Race and Category ?)

So returning back in the getTitleRace method :

  public getTitleRace(ranking: Ranking, playerIndex: number) {
    const playersInfo = ranking.playersInfo;
    if (!playersInfo) return "Random";
    const playerInfo = playersInfo[playerIndex];
    if (this.hasSelectedIcon(playerInfo)) {
      **return this.$t(`races.${ERaceEnum[playerInfo.selectedRace]}`);**
    } else {
     return this.$t(`races.${ERaceEnum[playerInfo.calculatedRace]}`);
    }
  }

Problem is RaceEnum has no value equals to 32 : 

export enum ERaceEnum {
 [..],
  TOTAL = 16,
}

Maybe the fix would just be just to add a new Value for the ERaceEnum equals to SPECIAL ?

cdm-ium:
This looks like it's fixed.

Cepheid-UK:
Fixed.

PR #227, opened 2020-10-26
#206: fix undefined title for users with non-race icons

SergeyKhval:


Cepheid-UK:
Checked - works, thanks. #206

PR #241, opened 2020-10-27
Address dates when banning players.

cdm-ium:
wip

See: https://github.com/w3champions/w3champions-ui/issues/219

### DIFF
--- a/public/env.js
+++ b/public/env.js
@@ -1,9 +1,9 @@
 window._env_ = {
-  // BASE_URL: "http://localhost:5000/",
-  BASE_URL: "https://statistic-service.w3champions.com/",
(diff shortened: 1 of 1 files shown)


public/env.js
```diff
@@ -1,9 +1,9 @@
 window._env_ = {
-  // BASE_URL: "http://localhost:5000/",
-  BASE_URL: "https://statistic-service.w3champions.com/",
+  BASE_URL: "http://localhost:35468/",
```
cdm-ium, line 2: Will change all this, just getting a stub pr up.

abeninski:
This is already fixed in https://github.com/w3champions/w3champions-ui/pull/267

Issue #224, opened 2020-10-26
Performance in last 10 games using match history


Cepheid-UK:
Either on match history or on the profile, it could be useful to show the player's form (this season) in their last 10 games,

e.g. WWLLWLWWLL

SergeyKhval:
opened PR #243

Cepheid-UK:
PR #243

PR #243, opened 2020-10-28
#224: add recent performance to profile page

SergeyKhval:


abeninski:
Can you put a screenshot in the PR pls

src/components/player/PlayerLeague.vue
```diff
@@ -133,6 +165,19 @@ export default class PlayerLeague extends Vue {
   get isRanked() {
     return this.modeStat.rank > 0;
   }
+
+  get lastTenMatchesPerformance(): string {
```
Cepheid-UK, line 169: Type needs to be `string[]`, otherwise it gives a type return error.

Cepheid-UK:
Looks great, can you make it support 2v2 AT as well please if possible?

2v2ATsupport

src/components/player/PlayerLeague.vue
SergeyKhval: Sure, will fix

SergeyKhval:
@Cepheid-UK API does not return data for 2vs2 AT. I am not quiet familiar with c#, so I wont touch stats service. If it's fixed, I guess recent performance will just work then

Cepheid-UK:
I realised after I posted that comment that the 2v2 "Recent Performance" includes AT games I believe?

I'm happy with how it is anyway. Nice work, I'll merge.

Issue #226, opened 2020-10-26
Add empirical cumulative distribution function on the MMR graph


alocate:
as helpstone suggested long ago, change the MMR graph to look like the following:

![cumulative](https://user-images.githubusercontent.com/19414319/97194682-bbf1d780-17aa-11eb-822a-d87bc5741ddb.png)

Remove the green bar on the current MMR graph and have a vertical line instead to indicate own MMR. This should be only frontend work.

jj-apps:
Does something like this look okay? Chart.js isn't easy to customize. I had to use a plugin to do this ([chartjs-plugin-annotation](https://github.com/chartjs/chartjs-plugin-annotation))

![image](https://user-images.githubusercontent.com/3699727/97278052-7b1fb000-17f6-11eb-8ece-b343ff5632af.png)

alocate:
hey, yeah that looks perfect for the green line - but still need the yellow curve :D to get the yellow line, you do a cumulative sum from the left to the right of the values in the bar chart, and divide by the total sum.

jj-apps:
@alocate How's this? Should the tooltip for the points on the yellow line show a percentage?

![image](https://user-images.githubusercontent.com/3699727/97355776-9faa7500-1854-11eb-8450-e1ebc81582a2.png)

alocate:
looks good! yeah the tooltip would make sense.

alocate:
cumulative* also. but ready for release i think!

jj-apps:
Any other changes to the labels or tooltips? If not I will submit a PR.

![image](https://user-images.githubusercontent.com/3699727/97367887-2405f380-1867-11eb-89d8-142e888ed3e4.png)

HelpstoneX:
This looks super nice. Well done!

HelpstoneX:
The only thing I could think of now is an additional y-axis on the right site for the cumulative curve (as it's also done in the screenshot Toxi posted)

alocate:
that'd be good, but I'm also fine without, since the tooltip gives the information anyway.

modmoto:
@jj-apps totally dig it =)

jj-apps:
@HelpstoneX Added the y-axis to the right.

![image](https://user-images.githubusercontent.com/3699727/97459762-57915e00-18f9-11eb-8838-357f4046e1f2.png)

jj-apps:
Another option for the toolitp would be "top X.x%" like on the profile page.

[...]

PR #246, opened 2020-10-29
Mmr cumulative graph 226

jj-apps:
#226


Cepheid-UK:
checked and this looks great.

reminder that this needs an "npm install" on deployment for the new dependency.

chartcumulative

cumulative

Issue #214, opened 2020-10-22
Performance degrades when opening match history, especially for 4v4


abeninski:
I notice that performance is degrading when match history is opened, especially for 4v4.

SergeyKhval:
Opened #249

Cepheid-UK:
#249 improves, any further improvement, make a new issue.

PR #249, opened 2020-10-29
#214: improve match history page performance

SergeyKhval:
Converting PlayerIcon to functional component and introducing a condition to render player's country flag improves performance on this page by approx 30-40%% on my machine (1.1s on prod vs 0.7s locally).

However there is another performance issue: tooltips used abundantly. I tried to render this page without any tooltips and it further improved performance by about 40% (0.7s with tooltips vs 0.4s without). Surely we cannot remove tooltips here, but refactoring them in a way that does not have such a big impact on performance looks like a big overhaul which will influence all UI and not just this one page. If I have some spare time I'll probably take a shot at it.

0.7s and 0.4s still looks very unoptimized, so may be it is better not to close the corresponding issue and investigate other options for improvements

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,8 +4,9 @@
   "private": true,
   "scripts": {
     "serve": "vue-cli-service serve --mode development",
-    "build:test": "vue-cli-service build --mode development",
-    "build:prod": "vue-cli-service build",
+    "serve:prod": "npm run build:prod && http-server ./dist",
+    "build:test": "rimraf ./dist && vue-cli-service build --mode development",
+    "build:prod": "rimraf ./dist && vue-cli-service build",
     "lint": "vue-cli-service lint",
     "format": "node_modules/.bin/prettier --write .",
     "update": "npm install -g npm-check-updates && ncu --upgrade && npm install "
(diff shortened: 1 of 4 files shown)


modmoto:
is this mergable? Or do you want to fix some stuff still?

SergeyKhval:
Yes, this is mergable. Just do not close corresponding issue

package.json
```diff
@@ -54,8 +55,10 @@
     "eslint": "^7.3.1",
     "eslint-plugin-prettier": "^3.1.4",
     "eslint-plugin-vue": "^6.2.2",
+    "http-server": "^0.12.3",
     "node-sass": "4.14.1",
     "prettier": "^2.0.5",
+    "rimraf": "^3.0.2",
```
abeninski, line 61: what is this for?

SergeyKhval: this is like a node alias for `rm -rf` command. performance testing is best done in production, so I added a local http-server to  host `dist` folder. however I stumbled upon some caching issues, so decided to remove dist folder each time before running build command

Issue #248, opened 2020-10-29
MMR Distribution graph does not show correct "Your MMR" for previous seasons.


Cepheid-UK:
@jj-apps great work on the MMR Distribution graph, could you update it so that the "Your MMR" line shows correctly for previous seasons too please?

jj-apps:
@Cepheid-UK Yes, will do.

Cepheid-UK:
PR #250 fixed.

PR #250, opened 2020-10-29
Reload "Your MMR" line when changing seasons

jj-apps:
#248


Cepheid-UK:
This fixed it, thanks! I'll merge.

I may create another issue for this as an enhancement, but the cumulative line has exposed how inaccurate the purple bars are, perhaps we can use the chartJS annotation plugin to make that a bit clearer, where a few percentile breaks are, 50/25/10/2 or something? Not sure, needs some thought.

Issue #201, opened 2020-10-22
Go-to-Player in rankings is inconsistent


Cepheid-UK:
Intended functionality: 
1. When a user searches a player name, the list will scroll to show the player.
2. When a user clicks on the division icon on a player's profile, the rankings will open and scroll to the player.

Issue:
1. This is inconsistent for even slightly less than full screen browser windows.
2. This does not work.

jj-apps:
@Cepheid-UK Are you able to reproduce the scroll issue on demand?

The only issue I see with scrolling is when going from the Grand Master page to a lower division with more than 30 ranks.

Cepheid-UK:
Tested and the functionality works, I'll check through to make sure I understand what you did for the future, then merge, thanks!

PR #251, opened 2020-10-30
Go to player in Rankings

jj-apps:
#201

### DIFF
--- a/src/components/ladder/RankingsGrid.vue
+++ b/src/components/ladder/RankingsGrid.vue
@@ -157,7 +157,7 @@ import CountryFlagExtended from "@/components/common/CountryFlagExtended.vue";
 import { ECountries } from "@/store/countries";
 import { TwitchStreamResponse } from "../../store/twitch/types";
 import RaceIcon from "@/components/player/RaceIcon.vue";
-import { getAvatarUrl } from '@/helpers/url-functions';
+import { getAvatarUrl } from "@/helpers/url-functions";
 
 @Component({
   components: {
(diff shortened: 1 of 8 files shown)


src/components/ladder/RankingsGrid.vue
```diff
@@ -357,15 +357,19 @@ export default class RankingsGrid extends Vue {
           behavior: "smooth",
         });
       }
-    }, 200);
+    }, 500);
```
jj-apps, line 360: @Cepheid-UK 

The scroll was occurring 0.2 seconds after the player was selected in the search box. This would fail sometimes when the rankings for the selected player's league hadn't finished refreshing. I increased the scroll delay to 0.5 seconds to give time for the rankings to finish refreshing from the API.

Cepheid-UK: It does seem like this will be a good change, but it's still not fixed the bug with not scrolling correctly :(

jj-apps: Let me troubleshoot this some more.

Did this one work for you? **"2. When a user clicks on the division icon on a player's profile, the rankings will open and scroll to the player"**

modmoto: this is still ongoing, right? I marked it as a draft, just to not merge this accidentally.

jj-apps: @modmoto  yes

jj-apps:
@Cepheid-UK I updated the PR to use Vuetify for the programmatic scrolling. Can you see if this resolves the issues you were seeing?

https://vuetifyjs.com/en/features/scrolling/

PR #253, opened 2020-10-31
252: use false in getAvatarUrl

echonax:


abeninski:
This will prevent classic icons from rendering. Not sure why you want to do this?

echonax:
I think i understood it the other way around, the request is about using the classics in the match history and player match history pages, my bad!

Issue #255, opened 2020-11-01
for finished match page for 4x4 / 2x2 use same players order for statistics as for players list


Qeenon:
example https://www.w3champions.com/match/5f9de61aef7438b9a583900c

SergeyKhval:
Opened #259

Cepheid-UK:
Fixed by #259

PR #259, opened 2020-11-01
#255: fix score orders for 2v2 and 4v4 matches

SergeyKhval:


Qeenon:
@SergeyKhval why there is `startsWith`, I think performance wise equality should work slightly faster, not important at all, it's just interesting for me why `startsWith` was used instead of equality here, maybe it's some js specific

SergeyKhval:
@Qeenon I didn't change part with `startsWith` in this PR, so better to ask the author of that code

Issue #242, opened 2020-10-27
Adding a dropdown menu for gamemode in the MMR distribution page


alocate:
Right now it only shows the MMR distribution for solo (and only europe? not sure). Would be sweet to have the same for other gamemodes, like some other pages in the Statistics section already have.

https://www.w3champions.com/OverallStatistics/mmr-distribution

modmoto:
afair it is bound to the gateway that you selected in the league overview.

jensk1:
I wanted to tackle this, however if i understand correctly, the API Service does only provide interface for season.
https://github.com/w3champions/w3champions-statistic-service/blob/master/W3ChampionsStatisticService/W3ChampionsStats/MmrDistribution/MmrDistributionHandler.cs#L17

This should also be extendended, to optionally filter by game mode. Also maybe mmr distribution per Race is interesting aswell? I heard humans are underpowered all the time ;). 

If possible i could contribute to statistics-service aswell, but i am not quiet sure how to get into it (and if this is a good first issue).

Cepheid-UK:
> I wanted to tackle this, however if i understand correctly, the API Service does only provide interface for season.
> https://github.com/w3champions/w3champions-statistic-service/blob/master/W3ChampionsStatisticService/W3ChampionsStats/MmrDistribution/MmrDistributionHandler.cs#L17
> 
> This should also be extendended, to optionally filter by game mode. Also maybe mmr distribution per Race is interesting aswell? I heard humans are underpowered all the time ;).
> 
> If possible i could contribute to statistics-service aswell, but i am not quiet sure how to get into it (and if this is a good first issue).

If you read up in #new-devs-chat on discord, abeninski has been discussing how to set up the stats service locally in order to make changes to it. He also provided some old data to use for testing. You also need to install mongodb, there is an outstanding task to update the stats-service with a wiki page that explains how to set up the dev environment. I'll make an issue there.

modmoto:
I will actually write up a small docu for that
@abeninski I will do the documentation right now.

jensk1:
@Cepheid-UK i got it working with the new docu.
But i cant find a #new-dev-chat in discord, is it internal?

jensk1:
I think this issue can be closed since it is live already. 
If there is still something to do let me know

modmoto:
jeah, u are right =)

PR #260, opened 2020-11-01
#242 adding dropdown to mmr

jensk1:
This is the implementation for feature https://github.com/w3champions/w3champions-ui/issues/242 
The Backend feature already has a Pull request in statistics service: https://github.com/w3champions/w3champions-statistic-service/pull/90

I reused the already existing components for the selections, as you can see the distribution changes for different selections:
1vs1
2vs2

America multiplayer looks kind of weird, but I think its just how it is: 
2vsUS

Nonetheless some one should code review this!
I had to create a new interface interface SeasonGameModeGateWayForMMR in Overallstatistics Type, because the actionhandling only allow one parameter Object (Payload). This is the wrapper for it.

If something needs to be changed, let me know.


jj-apps:
There is another dropdown for game already on the Profile/Match History screen that matches the same style:
image

modmoto:
jeah, that would be nicer, other than that, it seems solid.

jensk1:
I Changed it now to the dropdown. Some more code, but I agree it is more intuitiv.
Also I added that the default values will be display. Season 4 was selected by default but it was not shown.
Default gamemode ist 1v1 and shown now aswell. Default Gateway is Europe
dropdownPreselect

modmoto:
nice one, is this ready? I will update the test db aswell, so we have nicer data there.

jensk1:
From my side it is finished. If someone has reviewed it, you can go ahead :)

modmoto:
ok, lets see how this does on test =)

PR #263, opened 2020-11-02
remove condition to render player country flag

SergeyKhval:


modmoto:
i think this is actually needed to prevent errors, when no country is set, right?

SergeyKhval:
I just reverted my previous change here. the country-flag-extended component will decide by itself whether to use palyer's country code or location

Cepheid-UK:
infi

Not sure why Infi (or maybe china) is an exception to this...

Issue #219, opened 2020-10-25
Add date format validation to ban player screen


abeninski:
We need to add a validation for invalid date on the ban player admin screen.

cdm-ium:
I guess the problem date is the end date? Taking a look....

abeninski:
> I guess the problem date is the end date? Taking a look....

Yes.

SergeyKhval:
Opened #266

Cepheid-UK:
Did #266 fix the issue?

modmoto:
i will take a look during the overhaul stuff

gustav87:
This is nowadays only a problem if you insert a ban directly into the database. Could be improved by having the date in the db as a Date, not as a string, but I'm closing this as resolved, since there's not many inserting bans directly into the db, and they are currently aware of this.

PR #266, opened 2020-11-03
#219 ban end date

SergeyKhval:
Added date picker instead of plain text field

image


SergeyKhval:
Issues in git commits history, will open other PR

modmoto:
nice one =)

PR #269, opened 2020-11-03
search unranked players and route to profile page

knumf:


knumf:
connected to https://github.com/w3champions/w3champions-statistic-service/pull/91

src/views/Rankings.vue
```diff
@@ -203,6 +206,11 @@ export default class RankingsView extends Vue {
   @Watch("searchModel")
   public onSearchModelChanged(rank: Ranking) {
     if (!rank) return;
+
+    if (!this.playerIsRanked(rank)) {
```
modmoto, line 210: this will break on the profile page, when the player is not real. Also we have players that are unranken (have 2/3 games) that could be found somehow differently.

modmoto (commented):
see my comment on the stats service

src/views/Rankings.vue
knumf: the player will only be returned if it has a rank and/or playeroverallstats, then it should get battletag and routing to profile page should work. do you have an example i can use to break the profile page?

src/views/Rankings.vue
```diff
@@ -382,6 +390,16 @@ export default class RankingsView extends Vue {
   public async setLeague(league: number) {
     await this.$store.direct.dispatch.rankings.setLeague(league);
   }
+
+  public playerIsRanked(rank: Ranking): boolean {
+    return rank.player.games > 0        
+  }
+
+  public routeToProfilePage(playerId: string) {
```
abeninski, line 398: can you extract this in this file https://github.com/w3champions/w3champions-ui/blob/master/src/helpers/url-functions.ts

PR #276, opened 2020-11-07
Update Played Heroes Chart to show by percentage as well as change to a unique color scheme

JKoehler11:


JKoehler11:
Looks like someone made the same change I did to BarChart.vue. I will resolve this when I get back.

JKoehler11:
The percent is a bit weird on this one because it is missing played race. Currently, what comes back is { hero, count }. In the future, the micro-service will need to return { hero, race, count }

src/enums/Heroes.ts
```diff
@@ -0,0 +1,34 @@
+export enum HeroIndex {
+    Human = 0,
+    Orc = 1,
+    NightElf = 2,
+    Undead = 3,
+    Tavern = 4
+}
+
+export enum HeroNames {
```
modmoto, line 9: i would actually prefer having this in the backend as an auto property. You can calcualte this in the readmodel aswell.

modmoto (commented):
but that could be done pretty easy, tbh. Putting this in the backend would make the api nicer imo.

Cepheid-UK:
Is there more to do on this?

JKoehler11:
> Is there more to do on this?

Yes, I will try to get to this today.

Cepheid-UK:
nearly a one-year anniversary of this PR, I'm going to close it, because so much of the code/strategy has changed for this, if you'd like to resubmit it, please feel free, but we'll probably have to discuss it a bit via discord, on what the best way to proceed with this is. Feel free to @ me there. Some backend work is required here.

PR #280, opened 2020-11-12
New socials appearance

Cepheid-UK:
From:

support2

To:

support3


modmoto:
you saw that the pr build is red? some import seemst to be wrong

https://dev.azure.com/w3champions/w3champions/_build/results?buildId=3440&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9

modmoto:
And I actually think the first one looks nicer. Less text, more icons, right?

Cepheid-UK:
To be honest I'd actually like larger icons, but these seem to be the best we have.

I could make the entire thing smaller so that the icons are proportionally larger.

As for the error, it looks like the discord icon isn't being found? Very strange considering I took that screenshot with that path. No clue what's happening there, unless the icon isn't on the server?

Cepheid-UK:
Wait a second, is that error because I used forwardslash instead of backslash? 🤣

Cepheid-UK:
Apparently yes it was.

discordpath

Cepheid-UK:
Working with Riptide to produce some more appealing buttons, which may end up being useful in mutliple places on the site.

jj-apps:
> Working with Riptide to produce some more appealing buttons, which may end up being useful in mutliple places on the site.

I would make the "Find us on" and "support us" text a smaller font size or weight. That way the buttons/links will stand out more.

Cepheid-UK:
With Riptide's buttons, any feedback?

feedback

darkmodefeedback

mobilefeedback

PR #286, opened 2020-11-18
Player MMR Timeline

weidenbach:
This relates to issue https://github.com/w3champions/w3champions-ui/issues/215
Adds a player mmr timeline to the player profile
grafik

Please note that I changed the linechart.js to apply a gradient effect. Hence the linecharts on the statistics page will also have this effect. Imo this is nice, so I didn't bother changing it.
grafik

### DIFF
--- a/src/components/overal-statistics/LineChart.vue
+++ b/src/components/overal-statistics/LineChart.vue
@@ -8,6 +8,7 @@ import { ChartData } from "chart.js";
 })
 export default class LineChart extends Mixins(Line) {
   @Prop() public chartData!: ChartData;
+  @Prop() public beginYAxisAtZero?: boolean;
 
   private options = {
     legend: {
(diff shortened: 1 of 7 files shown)


modmoto:
@weidenbach 
I just saw that there was a build error in `LineChart.vue`. I just added a ts ignore so the build works again. Would be cool if you could fix this.

modmoto:
Just another thing: I switched the winrate and the stat tab. And did we not want to aggregate the winrates per day? On test it is still merged by game. And the style seems to be different than the other statistics, they are not rounded, or am I wrong?

weidenbach:
> 
> 
> @weidenbach
> I just saw that there was a build error in `LineChart.vue`. I just added a ts ignore so the build works again. Would be cool if you could fix this.

Added a PR for this https://github.com/w3champions/w3champions-ui/pull/290
Also adresses the line smoothing.

> 
> 
> Just another thing: I switched the winrate and the stat tab. 

I actually did this on purpose and asked Cepheid about his opinion about it. Imo the order doesn't really matter for both of them for long term, but for short term I thought it's better to place it at the top, just for adding visibility to a new feature. If someone doesn't know that the feature was added, it can be easily overlooked, as only the "headline" is visible (at least on my setup)
grafik
But yeah, for long term it doesn't make a difference. Also I guess an update note would also do.





> 
> 
> And did we not want to aggregate the winrates per day? On test it is still merged by game.

Yes, I didn't change it. I think it would be best to change this on the backend, to not send data that isn't needed. I can look into that the next days.

PR #290, opened 2020-11-20
Player MMR Timeline: Fixes LineChart.vue and adds rounding for Chart

weidenbach:
Fixes LineChart.vue gradient build error.
Adds PlayerMmrTimelineChart rounding for consistency with other charts.


modmoto:
just a quick question: will the backend do the aggregation of matches per day?

weidenbach:
Not yet, but I already programmed that and will create a PR today.

weidenbach:
I chose to take the highest mmr per day instead of an average or that last mmr, because it can be confusing when a player knows that his highest mmr was X but its not displayed in the graph.

modmoto:
not the last one? or average? The highest might lead to a distorted statistic, because you will have the most optimal statistic there.

weidenbach:
Hm, you made me reassess. I think you're right. While the highest is nice for the individual player, it doesn't depict the most accurate mmr. If we trust in the mmr system then the latest mmr should be the most accurate, as the system had more games to rate the player.

PR #292, opened 2020-11-22
Player MMR Timeline: Adds Rp, One MMR/RP pair per day, gradients refined

weidenbach:
A picture is worth a thousand words
grafik

grafik

This requires the PR https://github.com/w3champions/w3champions-statistic-service/pull/95 to be merged first

(This relates to https://github.com/w3champions/w3champions-ui/issues/215)


weidenbach:
One thing that is still not consistent between the line charts is how the dots appear.
grafik
I originally changed the dots to make the graph less cluttered, when it still had one entry per game. Now I could change it back, or we use the new design for the other line charts. I think especially the games per day chart could be less cluttered with smaller dots. What do you think @modmoto ?

abeninski:
This looks so sick. Good job mate!

modmoto:
Pretty great =D

PR #297, opened 2020-11-27
MMR/RP timeline Bugfix: Fixes bug that prevented timeline from loading if /player/statistics …

weidenbach:
Fixes bug that prevented timeline from loading if /player/statistics was directly loaded via URL. 

Also fixes the related bug for Matchup Statistics. On prod the correct tab is also not loaded.

The bug is that mounted calls of children are done before player.vue init is finished, so necessary data is not loaded in time.
The fix is adding an isInitialized state to player store, which is watched by children.

Directly loading via URL currently:
grafik

After fix:
grafik

Related to https://github.com/w3champions/w3champions-ui/issues/294


weidenbach:
Noticed a bug, will fix first

PR #298, opened 2020-11-29
rework admin page

modmoto:
things to do:
- use new chat ban rout of chat service
- color code ban types
- order bans by end time
- grey out bans that timed out

- use ban route of matchmaking service directly


modmoto:
thanks for marking as draft ;)

Cepheid-UK:
Merged #304 which is relevant to this.

PR #302, opened 2020-12-04
add ability to save link to trovo channel

SergeyKhval:


modmoto:
can we use their logo instead of a video icon?

SergeyKhval:
material design icons do not have one and I could not find svg online. may be we can ask them directly about their logo

PR #303, opened 2020-12-05
Shifts Popular Hours chart from UTC to local timezone, fixes bug that…

Boda805:
… caused 4 bars at beginning and one at end not to show. https://github.com/w3champions/w3champions-ui/issues/299 I'm having some confusion on what is indeed coming from the api in terms of time, and not sure if that has to do with running locally. If we get UTC time coming from api then this is basically ready, unless someone sees issues or things that can be improved.


src/components/overal-statistics/PopularGameTimeChart.vue
```diff
@@ -20,8 +20,8 @@ export default class PopularGameTimeChart extends Vue {
     const gameHour = this.popularGameHour;
     if (!gameHour) return [];
 
-    const times = gameHour?.playTimePerHour.slice(4);
-    times.pop();
```
Boda805, line 24: slice(4) caused bars 00:00, 00:15, 00:30, 00:45 and pop caused 23:45 to not go through

Cepheid-UK:
@Boda805 you mind rebasing this on top of master, if it's still relevant, I'll merge.

PR #305, opened 2020-12-07
add top live 1vs1 matches widget to homepage

SergeyKhval:


SergeyKhval:
Looks like this

image

modmoto:
oh wow, pretty cool idea!
is this done already?

abeninski:
Should we remove top 5 1v1 ladder and increase that list?

SergeyKhval:
@modmoto Yeah, just needs to be merged. Currently widget will show up to 5 1v1 games sorted by mmr desc. I can add other modes as well

modmoto:
@abeninski also a good idea. We have the top 5 when we go to the ranks. 
@SergeyKhval Does this show the live GM games? Or just the top 5 games overall? IT would be sad, if we had no GM playing and that list would be empty in off hours. It should reflect activity of pros or "not yet pros" =)

SergeyKhval:
It is 5 highest 1v1 mmr games overall. [Here](https://github.com/w3champions/w3champions-ui/pull/305/files#diff-4a700bb03f292f1ebac2afe2ccf7637fa07b23dd8702e860583cd008ce7411b9R61) I take 5 of them. We can show all active games, just the widget can get really tall. I can also add other game modes, just need to think about UI.

modmoto:
i released the stat service to prod, can you give me a heads up, when you tested the ui locally against prod?

modmoto:
just released the ui, seems to work =)

PR #308, opened 2020-12-10
Increase items per page on race->maps stats table.

### LINKED ISSUE #301: Autumn Leaves doesn't appear on statistics page, even when its' played
https://www.w3champions.com/player/iDentify%2311833/statistics

I suspect this is due to a capped # of maps being displayed? or perhaps its something else?

aaronjewell:
By default the Vuetify table component will only display 10 items
per page. This increases the limit to Number.POSITIVE_INFINITY
as it is very unlikely for the maps list to grow out of control
and any lower number would be arbitrary.

resolves #301

### DIFF
--- a/src/components/overal-statistics/RaceToMapStat.vue
+++ b/src/components/overal-statistics/RaceToMapStat.vue
@@ -3,6 +3,7 @@
     <v-data-table
       :headers="headers"
       :items="stats"
+      :items-per-page="Number.POSITIVE_INFINITY"
       hide-default-footer
       :mobile-breakpoint="400"
       :hidden="stats.length === 0"


aaronjewell:
Profile to demo the fix:

/player/Grubby%231278/statistics
season 4
patch 1.32.9

https://vuetifyjs.com/en/api/v-data-table/#items-per-page

Now it goes to 11!  (and more...)

Cepheid-UK:
Hey ty for the PR, I tested and it does resolve that issue so I will merge, but there is another case that doesn't seem to work:

This was the player who identified (Ah ha!) the issue and reported it:

This was a game he played on Autumn Leaves during S4:
https://www.w3champions.com/match/5fce9f06ef7438b9a5dbe56b

If you look at his S4 stats table, Autumn Leaves isn't there:
https://www.w3champions.com/player/iDentify%2311833/statistics

I checked with your PR locally and it doesn't fix that issue either. I'm not sure what's causing it, but seems like it was two issues 🤔.

aaronjewell:
Thanks for testing and merging @Cepheid-UK!

Checking on production, it does appear that iDentify's map list is now correct. I can see 11 Maps (including "Overall"), including Orc vs NE on Autumn Leaves with losses. I am not the most familiar with the site, so it is completely possible I am missing something :)

identify-map-list

Cepheid-UK:
That's a relief! I'll close the issue. Thanks again.

Issue #277, opened 2020-11-08
Add option to sort ongoing games and finished games by avg MMR and filter by map


HelpstoneX:
- makes it easier to spot high level games (to check livestreams or watch vods afterwards)
- can be used to find matches on a specific map to see how it gets played and which players don't have it vetoed (especially useful for mapmakers of the new maps)
- also very useful to check how well the matchmaking works on higher/ lower levels

jensk1:
Currently the api is only able to load the matches descending by ID of the match. 
If this feature is needed and approved I can extend the Statistics service, so it will be able to also return the matches sorted by map/mmr maybe duration.

modmoto:
jeah why not. Extending the search in the statistic service should be really straight forward. Maybe mmr is a little difficult, but mapname is easy.

jensk1:
You are right, map is pretty easy. MMR is kinda hard. Because MMR is structured in the Players and not directly in the Matchup Object. I thought best way is to query by Old MMR of all players participating in a match and average it. But this is not queriable by the Mongo DB driver. So maybe one possibility is to add the average MMR of a Game directly to the matchup in the DB, so it is queriable?

alocate:
another issue is that MMR is affected by resets - for another purpose (to computed expected winrates), we are thinking of adding a shadow MMR calculation which is not affected by resets (which would not be used anywhere else). this looks like we could kill two birds with one stone too, as we could also create an "average shadow MMR" for each match as the migration is done?

alocate:
if you want a more simple solution though, I think ordering games by MMR of the first player could work?

jensk1:
I will try it by ordering it by MMR of the first player, think this can be especially for 2v2 and 4v4 not the optimal ordering.
Does the MMR reset also affect the OldMMR attribute of Players in the matchup Collection? Think this is a snapshot of the Matchmaking. But still shadow MMR would be the better solution

Cepheid-UK:
Perhaps ordering by MMR is a job for the new python app?

modmoto:
we should let this in the stats service, as the mmr service should be used for business logic. I would not call two backends from the frontend just to do some sorting. I think it would be fine to call it asynchronously in the handlers, maybe, but not more than that. Just to keep the dependencies a little lower.

gustav87:
@Cepheid-UK Is this considered done you think?

PR #309, opened 2020-12-10
Add map filter to matches.

aaronjewell:
The match select uses a hard-coded list of maps. This might be
more appropriate to get from a API. I'm not 100% sure that the list
hard-coded in this commit is the most correct list of maps, but it
seems to be pretty close.

Front end implementation to address part of #277 
Companion back end PR at w3champions/w3champions-statistic-service#108


Cepheid-UK:
We really can't do that maps() function like that. It really has to be dynamically generated.

Look at this for some guidance:

https://github.com/w3champions/w3champions-ui/blob/2ea079324634d70161abcf6d608c6e6064dc423d/src/store/overallStats/index.ts#L58

aaronjewell:
I thought that might be the case. It now pulls the list of games by mode, and for finished matches it reduces the maps from all seasons, while ongoing matches are shown only current season maps played.

There are currently two outstanding issues, the first is that in some cases maps are translated into the same label: "Autumn Leaves", and there is at least one translation that doesn't exist for "lastrefuge.anon". That one will also need to be handled special because of the "." character in the name, I believe, will conflict with the translation method, since it uses dot notation for nesting. I haven't identified if there is a way to escape it in the translation module.

To handle the same name, I was considering appending the map key in parentheses on only duplicate labels, but the UX seems poor. I would love feedback on alternative ideas!

Cepheid-UK:
> "lastrefuge.anon"

That was actually a mistake during the S4 reset, we accidentally named one of the maps with a . instead of a _.

I believe the map file was renamed and any old entries in the db for "lastrefuge.anon" were migrated to the "lastrefuge_anon"

I don't _think_ that "lastrefuge.anon" will ever appear, and that translation should really be removed, but I'm not certain.

aaronjewell:
Great! I was using the test dump, so that makes sense. Perhaps putting `(Oct 20)` on the end of the Autumn Leaves translation for autumnsleaves_201016 is a decent tradeoff then since it's limited in scope to a single map?

Cepheid-UK:
Just bumping this because I've totally forgotten where this got to, is this ready for merge?

aaronjewell:
No worries! This doesn't solve the Autumn Leaves dilemma, but that's an existing data defect. This is ready, but needs the accompanying BE PR https://github.com/w3champions/website-backend/pull/108

Issue #212, opened 2020-10-22
End of season tournament profile badges


Cepheid-UK:
It would be nice to reward end-of-season tournament participants with a badge on the profile similar to the season participation badge.

Design as well as implementation is needed.

![badges](https://user-images.githubusercontent.com/40397054/96875703-71512200-146f-11eb-85b4-eccc7f6fb7b5.PNG)

jj-apps:
@Cepheid-UK Is there someone that could design the badge? Maybe something like the trophy icon in the menu? 
![image](https://user-images.githubusercontent.com/3699727/97529271-c6f46580-196c-11eb-9007-8608f47338b4.png)

Cepheid-UK:
Talk to Riptide on discord, he designed our badges.

jj-apps:
@Cepheid-UK 
![image](https://user-images.githubusercontent.com/3699727/98016146-dc2d2200-1dba-11eb-88df-d911c34a4a93.png)

Here is what Riptide came up with so far. 

He did ask if we wanted all participants to have a badge or just the top 3 (He made them Gold, Silver, and Bronze).

Cepheid-UK:
These look pretty good, want to take a crack at implementing them?

In an ideal world, we'd tie these to the admin or tournaments page so that admins can assign these, but that might be a bit wider scope than this specific issue.

jj-apps:
Yeah I'll give it a try.

He is going to create a 4th trophy for anyone that doesn't make the top 3, but was in the tourney.

modmoto:
The absolute ideal world would be that we get them from the tournamnet winner dynamically. But that would also include linking the tournaments with the actual battletag. Right now it is only a string. But this would save us some time and would prevent errors/forgetting to assign these.

Cepheid-UK:
made an issue #275 to cover this @modmoto it's a prerequisite for any further work on tournaments, and would help the admin page too. Nice suggestion.

modmoto:
nice one

jj-apps:
@Cepheid-UK @modmoto 

Updated images from Riptide:
![image](https://user-images.githubusercontent.com/3699727/98431546-89ca5a80-2073-11eb-9128-83af5586e60b.png)

There is a Gold, Silver, and Bronze trophy for 1st, 2nd, and 3rd place. All trophies have the season # on them.

He presented some choices for the rest of the tournament participants:
Option 1: Dark grey trophy
Option 2: Trophy colored based upon the "season color"

[...]

PR #321, opened 2020-12-31
[WIP] Tournament badges 212

jj-apps:
#212 Front end implementation of tournament badges

image
image


jj-apps:
@Cepheid-UK @abeninski 

Here's the front-end implementation of the tournament badges, but I need some guidance on how the backend should be built.

modmoto:
@jj-apps 
hey, I just saw that you searched for some guidance, 20 days ago ^^ 

What can I help you with? Did you built the backend allready or are you unfamiliar with .net?

modmoto:
I just realized in the backend, that we are having an issue with the tournament, that we should fix first. The tournament does not support battletags, just plain strings. So before we add the tournament badges, we should add an auto completion feature for the tournament badge and save the battletags within the tournament. Then we could update the player profile for the tournament aswell. And then the ui should work out of the box.

jj-apps:
> @jj-apps
> hey, I just saw that you searched for some guidance, 20 days ago ^^
> 
> What can I help you with? Did you built the backend allready or are you unfamiliar with .net?

@modmoto I'm familiar with .Net and can do the backend implementation for this. Just need guidance on where in the database this should be stored (first time using MongoDB for me).

Would adding BattleTag to TournamentPlayer be all we need? Then it can be returned via a join with the player profile?

I suppose we would need to update the Tournament admin UI to allow for entering the btag too.

modmoto:
jeah, updating the admin UI would be needed for that. But I would make it just a tiny bit more complex, in making the name field for the player a search box, like on the ranking page, that returns the small profile with name and btag. Then we can just take this battletag and name from the backend. This makes the tournament page much easier to handle and on top off that we can store the correct battletag in the backend. So we dont have to go through the old tournaments and searching the battletag by hand. I know this is a "little" more work than expected, but ultimately we wanted to do the search like that anyways and as we need that info, I think this is a good opportunity to go for it.

Regarding the mongodb:
Joining is a little weird, we are doing this on a few occasions but not very often. Like here for example: https://github.com/w3champions/website-backend/blob/master/W3ChampionsStatisticService/PersonalSettings/PersonalSettingsRepository.cs 
The important thing is the `Lookup<T>`. The dto has to have a list of some sort, where you join the data into. In a first and easy way, we could just load the tournaments when returning the profiles and do the join on our own. Performance wise, that is ok, as the whole DB is in Memory anyways. Codes wise it is a little rubish. But we have it like that for more complicated joins, for example profile and rankings. Another easy way would be to create a new collection with `TournamentResults` where you have an object like that 
```
{
  "battleTag": "peter#123",
  "participatedInTournaments": [
    {
       "id": "tour1"
       "place": 1
    },
    {
      "id": "tour2"
      "place": 3
    },
  ]
}
```
This can be easily written and more importantly we dont have to join. Like it is intended with mongo. We just load this, when someone opens a profile page and put it on the dto where you wanted it in this PR. We could also just make a separate route, to keep the objects separated (I would actually prefer that). Something like `api/tournament-results/{id}`. What I would not do, in any way, is to update the profile model with the tournament results. Because the profile is a readmodel and gets updated after every game. This means we can run into race conditions. To be fair, that will most likely not happen, because after we report a tournament result, the players are most likely not playing, but it is still a race condition that I would like to avoid. Keep the readmodels a read only collection for the syncers. 
Regarding technical stuff: There is a base repository in the backend that makes read and writes a little bit easier. But the `MongoClient` class is also not very complicated to use. When a collection is not present and you write to it, it gets generated automatically, so you dont have to have access to the db server for that. We have a open test db, that you can use (connectionstring is in the repo). I would recommend Mongodb Compass as a tool to access the db quickly and see if your data was stored right.

modmoto:
Just had a shower and scratch that idea with the additional collection. We can just add a route `api/tournament-results/{id}` that iterates over the torunaments and generates that dto. But we still need the battletag for that. I will make the API accept a battletag in the dto and add the route in a separate PR. Then you can focus on the UI for that.

abeninski:
I think after @gabcinder2004 tournaments get merged the winners for the next round will be auto determined so no more manual action from admin and no need to write the code to link battle tags?

modmoto:
But we still need battletags, right? Right now we only enter names and dont link them to the account.

abeninski:
People register from ingame and their battle tag is used

modmoto:
Also our season finals? Because this PR is about that.

abeninski:
Yes. Season finals should use Gab's tournaments as well. I don't see a reason why not use it?

modmoto:
yes, good idea. I was just wondering if this might be an issue with observers and stuff like that. Or if b2w and the other streamers ar getting this.

Cepheid-UK:
@jj-apps to say we have neglected PRs is an understatement 😄 but we're trying to get back on track.

This is still very relevant, we're going to be adding the ability for admins to give players badges, and the automated tournaments are still going to be using these too, somehow, not sure exactly yet.

do you want to update this and we can work on a backend solution, even if its only temporary or manually done.

PR #323, opened 2021-01-05
[WIP] Localize website

RickyMarou:
# WIP
---

This PR does a few things:

1. Get the locales from the [Google Sheet](https://docs.google.com/spreadsheets/d/1WCFB8n_DufVuPIS86Le33gKos6QIR99S0kLI_c-QMok/edit?pli=1#gid=0) and JSON serialize it to `src/locales/data.json` ( ✅ done )
2. Import the data to the `vue-i18n` plugin ( ⚪ TODO )
3. Replace all the hardcoded strings in the website ( ⚪ TODO )
4. Add a language selector in the UI ( ⚪ TODO, most likely will cherry pick from the [locale branch](https://github.com/w3champions/w3champions-ui/tree/locale) )

--- 
## Open questions/suggestions:

- When do we want to run the script for the locales (in CI or manually ?)
  - If we run it in CI we might want to check out the data out of versioning
- How do we manage keys ? We need some form of Auth to connect to the Google Sheet API (Im using an API Key). 
- The Google Sheet could be greatly improved by linking the individual worksheet for each language to the main sheet and the main sheet should be locked and made uneditable. The Google Sheet should not be publicly editable (publicly readable is OK imho), is there an org for w3c on Google ?


RickyMarou:
@Cepheid-UK @arcsaber interested in your feedback on the questions here, especially the one about credential management ☝️

Cepheid-UK:
Please look at the Locale branch, I've already implemented the language selector, can you submit work to that branch? No point duplicating.

I dont think for now we need a live updating feature from the doc, I think it's ok to periodically add these xy.ts files when we want to support a new language.

https://github.com/w3champions/w3champions-ui/tree/locale/src/locales

modmoto:
the keys can be stored as secrets in azure. But we can also just get the stuff during the build and put it as json in the build after the checkin, something like that. A buildstep in webpack or something like that. I would not do that manually, seems like a lot of annoying work ^^

RickyMarou:
@Cepheid-UK @modmoto thank you for the feedback.

I realised that the script still need some work as it should upsert and should allow people to create new keys for strings without using the google sheet if they need to, otherwise it makes introducing new features dependent on having the string localized in the Google Sheet. 

At this point I have a pretty good grasp on how to introduce the entire localization feature, im just short on time for the rest of the month between work and personal life but after that i'll be able to commit a decent amount of time to shipping it.

Is there a place for documentation for the w3c project ?

modmoto:
i would just put it here in the readme file, as it belongs to the ui.

RickyMarou:
Closing this PR now - will re-open from with the `locale` branch once it's closer to completion.
You can check the progress in my fork ( https://github.com/RickyMarou/w3champions-ui/pull/1 )

Issue #202, opened 2020-10-22
Random players icons do not show consistently in multiple locations


Cepheid-UK:
Random players Icons do not show correctly in the following places:
- Rankings page
- Live game component
- Match history

![infi1](https://user-images.githubusercontent.com/40397054/96870965-8cb92e80-1469-11eb-9523-4451db0db9b6.PNG)

![infi2](https://user-images.githubusercontent.com/40397054/96870976-8f1b8880-1469-11eb-8a95-fd0acba97d43.PNG)

![infi3](https://user-images.githubusercontent.com/40397054/96870986-92167900-1469-11eb-9ebe-ff8670b07436.PNG)

marco93f:
There's nothing wrong in showing the Reforged mud golem icon for live games and in match history, but that should not be the case for the rankings page. Just look at IFFI#1561, for instance: on his profile he has the Reforged sheep icon, but in the rankings page he still has the mud golem. You always want to show the profile icon.

Cepheid-UK:
Sheep is the default icon if the player doesn't select any other, right? 

Should the mud-golem in the rankings/matches grid be switched to a sheep if the player hasn't selected an icon?

Cepheid-UK:
Did a6ae5af fix this?

PR #326, opened 2021-01-06
fix issue #202 - profile picture correctly showing in /Rankings page

AscaL:
Simple change. Just put the 'usual' way of getting the picture. 
Is it expected behaviour that the function `getAvatarUrl` for `selectedRace = 16` and `pictureId = 0` (unselected) returns [TOTAL_0](https://storage.googleapis.com/w3champions-prod/integration/icons/raceAvatars/TOTAL_0.jpg) while getAvatarUrl for `selectedRace = 0` and `pictureId = 0` (sheep selected) returns [RANDOM_0](https://storage.googleapis.com/w3champions-prod/integration/icons/raceAvatars/RANDOM_0.jpg) , that in the end are the same?
If so this should be the correct solution i believe.

Cheers :)

### DIFF
--- a/src/components/ladder/RankingsGrid.vue
+++ b/src/components/ladder/RankingsGrid.vue
@@ -358,7 +358,12 @@ export default class RankingsGrid extends Vue {
         playerInfo.isClassicPicture
       );
     } else {
-      return this.raceIcon(playerInfo.calculatedRace);
+      return getAvatarUrl(
+        playerInfo.selectedRace,
+        playerInfo.pictureId,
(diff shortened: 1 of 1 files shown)


abeninski:
If there is no selected race this will still work?

AscaL:
Yes, what happens is that if it's not selected, the fn `getAvatarUrl` gets called with parameters `selectedRace = 16` and `pictureId = 0` and returns the TOTAL_0 image, that is a sheep.

Issue #208, opened 2020-10-22
Show rankings by country


abeninski:
It would be awesome if we could see rankings by country.

Example UI:
1) User clicks on a flag in rankings page and is redirected to a new page.
2) New page has rankings grid sorted descending by Grand Master players on top and Bronze players on the bottom. If there are no Grand Master player then Master is first and so on.
3) Leagues are separated with league name and title.

**Grand Master**
1. Rank 1
2. Rank 2

**Master**
1. Rank 1
2. Rank 2
3. Rank 3

**Diamond 1**
1. Rank 1
2. Rank 2
3. Rank 3

**Diamond 2**
1. Rank 1
2. Rank 2
3. Rank 3

JKoehler11:
Since we already have their country on the rankings page, I could see this as just a country filter.

I'll take a look into this quick to see how feasible this is.

Cepheid-UK:
Refrogged actually do this for French players and their implementation is how Abe described it.

![refrogged](https://user-images.githubusercontent.com/40397054/99060090-6f264300-2597-11eb-9d2a-aecb485d0529.PNG)

abeninski:
They do it using my tool 👍

jj-apps:
Not sure if anyone is still working on this. But since it is widely requested I'm going to start working on it now.

modmoto:
nice one =)

jj-apps:
Something like this okay?
![image](https://user-images.githubusercontent.com/3699727/103959325-9ec37880-510c-11eb-88e0-be2ef9fc23f2.png)

abeninski:
Just what I intended it should look. Great!
Are you separating by divisions?

PR #327, opened 2021-01-08
Add country rankings page

jj-apps:
Accessed by clicking on a player's flag on the Rankings page:
image

### DIFF
--- a/src/components/ladder/CountryRankingsGrid.vue
+++ b/src/components/ladder/CountryRankingsGrid.vue
@@ -1,0 +1,498 @@
+<template>
+  <div class="custom-table-wrapper elevation-1">
+    <table
+      class="custom-table"
+      v-for="countryRank in rankings"
+      :key="countryRank.league"
+    >
+      <thead>
+        <tr>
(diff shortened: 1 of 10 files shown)


jj-apps:
#208

jj-apps:
statistics service PR: https://github.com/w3champions/w3champions-statistic-service/pull/114

jj-apps:
I think this is ready now:
image

src/components/ladder/CountryRankingsGrid.vue
```diff
@@ -0,0 +1,535 @@
+<template>
+  <div class="custom-table-wrapper elevation-1">
+    <table
```
abeninski, line 3: Is there no way to use RankingsGrid component here instead of copying the whole logic?
https://github.com/w3champions/w3champions-ui/blob/master/src/components/ladder/RankingsGrid.vue

abeninski:
Mergin this because it's too cool to sit in PR. Hopefully @jj-apps will refactor it later

Issue #317, opened 2020-12-21
Some match summary pages are not displaying


Cepheid-UK:
Example:

https://www.w3champions.com/match/5fe09f70e8fc1825175cfb8c

Match history of iDentify#11833 (who reported the issue) seems to have a disproportionate number of these:

https://www.w3champions.com/player/iDentify%2311833/matches

The data appears to be received correctly, maybe an error in the client rendering.

![concealedhill](https://user-images.githubusercontent.com/40397054/102787378-e7530600-4398-11eb-9958-8f7e76cb7e84.PNG)

jj-apps:
@Cepheid-UK 
It seems to be only happening for iDentify#11833 in 1v1 flo games where he was the winner.

The error is caused by his battletag being returned from the api without the #number in the playerscores array.

Returns as `'iDentify'`, but should be `iDentify#11833`

jj-apps:
@Cepheid-UK Can we close this now?

modmoto:
seems fine to me, right? I close it. I you @Cepheid-UK has anything against it, we can reopen =)

Thanks for the fix! @jj-apps

PR #328, opened 2021-01-09
Match summary pages not displaying

jj-apps:
#317 
Add logic to handle incomplete battle tags. The issue is caused by incorrect battletag data stored in the database (i.e. 'iDentify' is in the db, but it should be `iDentify#11833`). The new logic will attempt to match the player if no matching battletag is found.

[w3champions/w3champions-statistic-service/#113](https://github.com/w3champions/w3champions-statistic-service/issues/113)

### DIFF
--- a/src/views/MatchDetail.vue
+++ b/src/views/MatchDetail.vue
@@ -333,47 +333,36 @@ export default class MatchDetailView extends Vue {
     await this.$store.direct.dispatch.matches.loadMatchDetail(this.matchId);
   }
 
-  private getPlayerScores(team: Team) : PlayerScore[] {
+  private getPlayerScores(team: Team): PlayerScore[] {
     let scores: PlayerScore[] = this.playerScores
       .filter((s) =>
-        team.players.some((player) =>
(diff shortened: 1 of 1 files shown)


src/views/MatchDetail.vue
```diff
@@ -272,12 +285,45 @@ export default class MatchDetailView extends Vue {
...
-        this.match.teams[0].players.some((player) =>
+    const winningTeam = this.match.teams[0];
+    let scoresOfWinners = this.playerScores
+      .filter((s) =>
+        winningTeam.players.some((player) =>
           player.battleTag.startsWith(s.battleTag)
         )
-    );
+      )
+      .map((s) => {
```
abeninski, line 295: Can you extract this into a function with some parameters cuz it seems like alot of duplicated code:
full/partial battleTag match, winner/loser

jj-apps: @abeninski PR has been updated.

abeninski: Oh much better

PR #330, opened 2021-01-10
WIP Add locales

RickyMarou:
🚧 Work in progress

✅ Pulling data from the Google Sheet  
✅ Injecting it to the `vue-i18n` as default/fallback values  
IN PROGRESS: 🕵️ Check all the strings in the website (😭😭😭)
TODO: 💄 Move the language selection dropdown to it's own component
TODO: 🕵️ Thorough QA

## Progress tracking:

```
src/views
├── Admin.vue ✅
├── Faq.vue ✅
├── GettingStarted.vue ✅
├── Home.vue ✅
├── Imprint.vue ✅
├── Login.vue ✅
├── MatchDetail.vue ✅
├── Matches.vue ✅
├── OverallStatistics.vue ✅
├── Player.vue ✅
├── Rankings.vue ✅
└── Tournaments.vue ✅



src/components
├── admin ✅
│   ├── AdminBannedPlayers.vue ✅
│   ├── AdminLoadingScreenTips.vue ✅
│   ├── AdminNavigation.vue ✅
│   └── AdminNewsForLauncher.vue ✅
├── clans ✅
│   ├── AcceptInvitePanel.vue ✅
│   ├── ClanCreationPanel.vue ✅ 
│   ├── ClanOverview.vue ✅
│   ├── ClanRoleIcon.vue ✅
│   ├── DeleteClanModal.vue ✅
│   ├── InvitePlayerModal.vue ✅
│   ├── LeaveClanModal.vue ✅
│   ├── MemberManagementMenu.vue ✅
│   └── PendingInvitesPanel.vue ✅
├── common ✅
│   ├── CountryFlagExtended.vue ✅
│   ├── GameModeSelect.vue ✅
│   ├── GatewaySelect.vue ✅
│   ├── LocaleIcon.vue ✅
│   ├── SocialBox.vue ✅
│   └── SupportBox.vue ✅
├── ladder ✅
│   ├── CountryRankingsGrid.vue ✅
│   ├── LeagueIcon.vue ✅
│   ├── PlayerRankInfo.vue ✅
│   ├── RankingsGrid.vue ✅
│   ├── RankingsRaceDistribution.vue ✅
│   └── SwordIcon.vue ✅
├── match-details ✅
│   ├── HeroIcon.vue ✅
│   ├── HeroPicture.vue ✅
│   ├── MatchDetailHeroRow.vue ✅
│   ├── MatchHiglights.vue ✅
│   └── PlayerPerformanceOnMatch.vue ✅
├── matches
│   ├── HostIcon.vue
│   ├── MatchesGrid.vue
│   ├── MatchesStatusSelect.vue
│   ├── PlayerIcon.vue
│   ├── PlayerMatchInfo.vue
│   ├── StreamedMatchInfo.vue
│   ├── StreamedMatchPlayerInfo.vue
│   ├── TeamMatchInfo.vue
│   └── TopOngoingMatchesWithStreams.vue
├── overal-statistics
│   ├── AmountPerDayChart.vue
│   ├── AmountPerGatewayPerDayChart.vue
│   ├── BarChart.vue
│   ├── GameLengthChart.vue
│   ├── HeroPictureSelect.vue
│   ├── HeroWinrate.vue
│   ├── LineChart.vue
│   ├── MapsPerSeasonChart.vue
│   ├── MmrDistributionChart.vue
│   ├── MultipleAmountPerDayChart.vue
│   ├── PlayedHeroesChart.vue
│   ├── PopularGameTimeChart.vue
│   ├── RaceToMapStat.vue
│   └── tabs
│       ├── HeroTab.vue
│       ├── MmrDistributionTab.vue
│       ├── PlayerActivityTab.vue
│       └── WinratesTab.vue
├── player
│   ├── ModeStatsGrid.vue
│   ├── PlayerAvatar.vue
│   ├── PlayerLeague.vue
│   ├── PlayerMmrRpTimelineChart.vue
│   ├── PlayerStatsRaceVersusRaceOnMap.vue
│   ├── PlayerStatsRaceVersusRaceOnMapTableCell.vue
│   ├── RaceIcon.vue
│   ├── RecentPerformance.vue
│   ├── SeasonBadge.vue
│   ├── TrovoIcon.vue
│   └── tabs
│       ├── PlayerArrangedTeamsTab.vue
│       ├── PlayerMatchesTab.vue
│       ├── PlayerProfileTab.vue
│       └── PlayerStatisticTab.vue
└── tournaments
    ├── TournamentBracket.vue
    ├── TournamentMatch.vue
    ├── TournamentMatchUpdate.vue
    ├── TournamentRoundConnector.vue
    ├── TournamentSelect.vue
    ├── TournamentStraightOpenConnector.vue
    ├── TournamentStraightOpenDownConnector.vue
    └── TournamentYConnector.vue

```


RickyMarou:
image

Not so sure why GH is showing these commits as part of this PR as they are already on master with the same commit hashes 🤔

modmoto:
Hey, maybe we should discuss how to incorporate this into the build? Or maybe the release? I would prefer build actually, so every release stays the same.

modmoto:
I think it would be allready buildable, if you add `generate-locales &&` on the build command in the front. For the environment we can add a buildstep to add the env variable as a token file. Then we do have the google secret in azure and every build pulls the new translations. 

I was also thinking of doing this on a container restart. Like in the `start.sh` script. That would actually be cooler in that regard, that we can rollout translations much more quicker and code independant by just restarting the container.

Let me try something locally and add a PR for your PR.

abeninski:
This sounds like sit lots of work. Why not just create files for each translation and release new ver?

modmoto:
only because we are blocked on translations when we have code that we can not release yet. I have a poc ready, but can not test it because i dont have docker on this machine. But the buildstep is absolutely fine for the first approach.

modmoto:
https://github.com/w3champions/w3champions-ui/tree/add-translations-on-restart 
I was thinking of something like that, could not test it yet, though.

Cepheid-UK:
Great work. I think translating on build is a great idea. Makes maintaining a lot easier.

Cepheid-UK:
> image
> 
> Not so sure why GH is showing these commits as part of this PR as they are already on master with the same commit hashes 🤔

Because since I last merged master into this branch, there's been some additional changes to master haha.

I've re-merged master into locale.

RickyMarou:
@modmoto your current version

> [`add-translations-on-restart` (compare)](https://github.com/w3champions/w3champions-ui/compare/add-translations-on-restart)
> I was thinking of something like that, could not test it yet, though.

I checked the code it is looking pretty good yeah 👍 
I also think running the script when starting the container is a pretty good idea, this way if we need to update the content from the GSHeet then we can restart the app instead of doing a new deploy.

I'll update the description with a better description when im closer to release.
We will absolutely need to add proper access control to the GSheet, otherwise anybody can deface the website by editing it 😅 

Also the scope of this PR will be strictly limited to static content, i do want to explore some solutions for localising dynamic content later tho, let's see.

RickyMarou:
@Cepheid-UK i just realised my PR is targeting the locale branch instead of `master` - this is why the commits from master are showing up in the diff 😞

modmoto:
> I also think running the script when starting the container is a pretty good idea, this way if we need to update the content from the GSHeet then we can restart the app instead of doing a new deploy.

and especially we do not have to create a new image. So we do not block ourselfes if we just want to correct a spelling mistake. Downside to this is of course, that the content of the website is kind of fluent and not related to the last release. Which is no big downside, imo. Another bigger downsid is, that we can make the container unstartable, if the script for creating the translations fails. So the script might need some roughing out, so it can handle errors in the translations. It would be bad if the container can not be started anymore because the translation has a missing key for example. Maybe a try catch for the translations and then we just add the english ones as default. Don`t know, something like that. We should definitely test what happens when the google secret is wrong or if the excel structure changes by accident, for example.

For the first approach we can just do it in the build, because it is really easy. But if you have the time, the restart approach is obviously my favoured one =)

RickyMarou:
@modmoto yeah let's go with the easiest solution first.
Actually after some thinking I think build time is the best because we can do things slightly differently than what you are doing currently in your branch:

It would be better to not write the locales to a separate JS file, because it makes that JS file render blocking and thus affects performance quite a bit.
It would be best to generate the data before the build and let webpack bundle the file with the rest of the app. 

Another big argument against using a standalone JS file is cache busting:
Currently all JS script are cached on cloudflare 👇 
image

Of course it's not a big deal to go the cloudflare dashboard and add a bypass rule for that one specific file, but we have to keep in mind that this will most likely be an ever growing file. It is not good to both disable cache on it and make it render blocking.

We don't have any of these problems if we let webpack bundle the data with the app as I can see there is already fingerprinting on the JS assets.

modmoto:
Writing to the js file would happen before the container is up or before nginx is started in the container to be specific. This load time would only occur once on startup. And yes, you are right, the cloudflare chache is a problem for that aswell. But we should actually put the translations in cloudflare aswell, they are 1,5k lines allready with only 3 languages. This might get big quite fast. We could still generate it during the build and add a release step to refresh the translations after the service is live. I wanted to do that for the ingame.js aswell, so figuring out how we do that would be reusable. But this is also something we do not need right now aswell. But I would really vote for at least the creation of the languages within the build. Maybe because I have my devops ocds ;)

Can you push the branch to this repo? Because I am too dump to clone your repo and make a pr for this branch xD

RickyMarou:
@modmoto still cannot push to this repository - can you double check my permissions please ?

image

modmoto:
can you try now? You actually should have write access, so merging should also be possible.

RickyMarou:
@modmoto yes it's working now i can push to this repo. 
@Cepheid-UK is it ok to force push all of the commits to the `locale` branch on the upstream repo ? I can also create a different one if you prefer

RickyMarou:
Re-opened with a branch not from my fork but from the same repo instead: https://github.com/w3champions/w3champions-ui/pull/335

PR #332, opened 2021-01-11
Fix match details for Flo matches

Qeenon:
players battletag include numbers after #

in playerScores those are missing

so it should work if we check if first starts with second (not other way)

image

image

### DIFF
--- a/src/views/MatchDetail.vue
+++ b/src/views/MatchDetail.vue
@@ -275,7 +275,7 @@ export default class MatchDetailView extends Vue {
     const scoresOfWinners = this.playerScores.filter(
       (s) =>
         this.match.teams[0].players.some((player) =>
-          player.battleTag.startsWith(s.battleTag)
+          s.battleTag.startsWith(player.battleTag)
         )
     );
     const scoresOfWinnersByBattleTag = _keyBy(scoresOfWinners, "battleTag");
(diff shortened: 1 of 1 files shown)


jj-apps:
@Qeenon this doesn't work for this match:
https://www.w3champions.com/match/5ff2f2abe8fc182517bdb079

I also already have a PR for this issue #328 . Feel free to suggest any changes to it.

Qeenon:
@jj-apps sry, hadn't checked if that already in PRs but may you explain why my solution will not work?

just in case of interest, for first example `ENA1337#2702` starts with `ENA1337` ?

jj-apps:
> @jj-apps sry, hadn't checked if that already in PRs but may you explain why my solution will not work?
> 
> just in case of interest, for first example `ENA1337#2702` starts with `ENA1337` ?
@Qeenon 
Well I ran your code and tried to open those matches and it still didn't work. I'd recommend that first. If you need help getting the project running locally against the api let us know.

It's not always just missing the #1234 portion of the battletag missing. In the second match example I gave you the battletag is `ichiban#1251` but it returns `BFoGG.Ichiban` instead. That scenario isn't handled by your code.

In the `ena1337` example it still doesn't display the match detail because of the `_keyBy(scoresOfWinners, "battleTag");` code. Check the result of that and you'll see what I mean.

Let me know if that helps.

Qeenon:
@jj-apps yes, thank you for explanation.

PR #340, opened 2021-01-17
Bugfix: Remove inactive game modes from games per day chart in Statistics page

RickyMarou:
☝️ What the title says 

Before:

image

After:

image

We most likely have the same problem across the website in multiple places.
I would suggest filtering out these game modes from the response that the backend gives, or for a pure frontend fix we could have a util `getGameModes` somewhere but i would like to stay for from the state management for the foreseeable future 😇

### DIFF
--- a/src/components/overal-statistics/MultipleAmountPerDayChart.vue
+++ b/src/components/overal-statistics/MultipleAmountPerDayChart.vue
@@ -34,9 +34,6 @@ export default class MultipleAmountPerDayChart extends Vue {
           return c.gameMode !== 9 && c.gameMode !== 6;
         })
         .map((c) => {
-          console.log("EGameMode[c.gameMode]", EGameMode[c.gameMode]);
-          console.log("c.gameMode", c.gameMode);
-
           return {
             label: this.$t(`gameModes.${EGameMode[c.gameMode]}`).toString(),
             data: c.gameDays
--- a/src/components/overal-statistics/MultipleAmountPerDayChart.vue
+++ b/src/components/overal-statistics/MultipleAmountPerDayChart.vue
@@ -27,17 +27,26 @@ export default class MultipleAmountPerDayChart extends Vue {
   get gameHourChartData(): ChartData {
     return {
       labels: this.gameDayDates,
-      datasets: this.gameDays.map((c) => {
-        return {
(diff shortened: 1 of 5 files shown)


src/components/overal-statistics/MultipleAmountPerDayChart.vue
```diff
@@ -27,17 +27,26 @@ export default class MultipleAmountPerDayChart extends Vue {
...
-        };
-      }),
+      datasets: this.gameDays
+        .filter((c) => {
+          // Filter out gameMode 9 as it does not exist
+          // and gameMode 6 as it is 2v2 AT which has been merged with 2v2 RT
+          return c.gameMode !== 9 && c.gameMode !== 6;
+        })
+        .map((c) => {
+          console.log("EGameMode[c.gameMode]", EGameMode[c.gameMode]);
```
gabcinder2004, line 37: Probably don't need these comments anymore?

gabcinder2004 (commented):
Overall looks good to me. Consider dropping the console.logs though. :)

modmoto:
is game modes undefined custom games? Would be cool in showing them aswell, right?

src/components/overal-statistics/MultipleAmountPerDayChart.vue
modmoto: +1

RickyMarou: Nice catch i let those slip through during my cleanup.

RickyMarou:
> is game modes undefined custom games? Would be cool in showing them aswell, right?

I was wondering the same thing. Anybody can confirm that ? Then I would add it to the enum and show it.

Would be clearer if the backend was returning strings instead of ints for game modes 😬
Imagine if in the future there is footmen frenzy, dota, etc... it'll be hard to keep track of what's what.

RickyMarou:
@modmoto what about merging like this and enabling when custom games stats are available ?
image

modmoto:
Jeah, seems fine, thanks for the pr :)

modmoto:
Hmm, don't find the merge button on mobile, will merge when I am back home.

modmoto (approved):
Ah here it is, after the review :D

Issue #295, opened 2020-11-25
Interface in the Admin page to add News


Cepheid-UK:
Modmoto moved the news items to the database, so we can now support updating the news through the website admin page.

I know Warcraft3.info uses TinyMCE to write WYSIWYG HTML, might be worth using too?

I also wonder if storing the news in raw HTML rather than string text might have some advantages, e.g. adding images, video embeds, fancy stuff to news posts?

modmoto:
We had html earlier but none coders should be able to do that aswell. Right now it supports markdown, so technically images aswell. But this might get weird when we have to scale them for different sizes. Also we need a Upload function for images aswell unless we reference from the static content but then it gets more complicated for floss and other non techies

Cepheid-UK:
Agree on uploading images. I think Abeninski had a good hosting solution for the in-game assets, so using that would make implementing anything that we want on both webstite + in-game easier.

aaronjewell:
I suggest tiptap for the text editor. There is also a Vuetify version that will integrate nicely with the site theme (although this is just admin facing anyway...)

It will take a URL for image inserts, and putting the images on a CDN is a great choice. Perhaps a separate image upload Vue component would make sense so you can achieve the dual purpose @Cepheid-UK mentions for in-game + website assets.

Cepheid-UK:
That does sound ideal, I think resolving the whole admin functions and such will be necessary before we can get to the 2nd phase of auto-tours, which will probably be our next priority after S6 launches.

Cepheid-UK:
I'm going to close this issue since this feature was added now. I'll open another if we have more specific feature requests.

aaronjewell:
@Cepheid-UK - Going to resurrect this, as I had some time to try out a WYSIWYG implementation that turned out to work pretty well. Let me know if you want to move forward with it.

PR #347

PR #347, opened 2021-01-25
WYSIWYG for admin news and simpler race theming.

aaronjewell:
The current editor I believe supports Markdown, submitting this PR as a possible alternative. This also includes some theme enhancements to make future styling simpler and support more components "out of the box" from Vuetify.

Assuming I understand the Azure auto-deploy correctly... [PR Deployment](https://pr.add-admin-news-content.ui.w3champions.com)

- Add WYSIWYG editor to the admin news create/edit
- Use variables to theme Vuetify. Enhance theme switcher using CSS variables + theme SCSS variables to allow for variations of dark/light based on race. Greatly reduces the number of color overrides and `!important` tags in the styles. Also allows more components to be themed automatically "out of the box."

Minor tweaks
- Some tab navigation was missing "exact" attributes to properly highlight selected
- Switched some inline transparent styles to theme classes.

wysiwg-editor


aaronjewell:
Implements #295 with an alternative to markdown editing.

modmoto:
Pretty cool =D

Regarding the PR: It looks like the mechanism does not work for PRs outside the org, because you are missing secrets in the build. That is why the build is also red. I can just add you as a new-dev to the team and you push the branch to the repo directly?

modmoto:
Also: does this create md content? Because the launcher treats the news as md content and renders it in html. So it would be good to check if we need changes in the launcher for that aswell.

modmoto:
ok, i added you with a new-dev role. That should make website and website-backend writable for you. 

Regarding the md in the launcher: This is also just a div with content. I guess we can use the same render methoid like on the website, as it is an electron app. Building the launcher is somewhat tricky, so if you have issues with that, just reach out. I could also do it, if it is just about adding a npm install and a new component. But we can check that later.

modmoto:
I am not emotionally attached to md, we just had it in place of a WYSIWYG  for poor people. But the news still have to show up in the launcher somewhat properly.

aaronjewell:
I like the idea of the launcher being able to handle markdown and html at the same time. We can leave the vue-markdown component, and v-if/else it with a regular v-html div. Should be easy enough to check. I'm thinking looking at the starting and ending character of the whitespace trimmed news content item and checking for `<` and `>`. That should be enough to distinguish it from md.

aaronjewell:
PR for launcher update at w3champions/launcher#80

aaronjewell:
Moving PR to organization account #351.

PR #351, opened 2021-01-29
WYSIWYG for admin news and simpler race theming.

aaronjewell:
Carryover from #347 under organization so build checks and run.


modmoto:
and the deployment worked =D

https://admin-news-content.ui.w3champions.com

aaronjewell:
🕺 🎈 🎉

modmoto:
We should wait until the launcher is released with the PR you made there. Then I am happy to do that =D

aaronjewell:
@modmoto - Should be good after the last couple releases!

modmoto:
ah jeah right, you want to resolve the merge conflict and merge from master again? Then we can just release. Backend is already out and fine, right?

aaronjewell:
Yep! Uses existing endpoints for news item with same data structure. Only potential defect that might arise is a race theme colored element being incorrect due to the theming enhancements. The ability for the launcher to display HTML is live.

PR #352, opened 2021-01-30
removing an scss override that was causing v-btn to not show "current" state

cmavelis:
News carousel, nav tabs, etc.  Looking at the blame, @kato88 added this as a bulk change, so I expect it wasn't intentional.  If anyone knows of a button where we _don't_ want this, it would be good to know

Screen Shot 2021-01-30 at 12 50 29 PM
Screen Shot 2021-01-30 at 12 50 54 PM

### DIFF
--- a/src/scss/shared/overrides.scss
+++ b/src/scss/shared/overrides.scss
@@ -33,10 +33,6 @@ h3 {
   font-family: "Friz Quadrata Std Bold";
 }
 
-.v-btn:before {
-  background-color: transparent !important;
-}
-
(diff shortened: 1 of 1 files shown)


cmavelis:
see: https://vuetifyjs.com/en/components/carousels/
for default styling example

modmoto:
ah you mean the grey outline? jeah cool, will merge

modmoto:
released on prod, thanks =D

cmavelis:
JOBS DONE

PR #355, opened 2021-02-01
minor improvements to player page

joleb:


modmoto:
I thought I deleted the package lock and moved it to gitignore? You want to expose that again?

joleb:
> Thanks for the pr!
> 
> I thought I deleted the package lock and moved it to gitignore? You want to expose that again? Better make it with ncu, that makes it clearer.
> 
> Rest seems to be fine :)

ok 👍🏻 shouldnt package-lock always be committed? But i can remove it for sure :P

modmoto:
we have a ^version thing so we always take the exact version and dont use package lock. That makes it less confusing, imo

PR #389, opened 2021-05-24
getting rid of most eslint warnings

joleb:
could not get them all, because I was not really able to recreate the intentions behind of some decisions.
I guess it makes also sense to upgrade "tiptap", once v2 is ready. It has built in typescript support and it looks much cleaner than the stuff we are doing :D


Cepheid-UK:
@joleb this ones up next, needs some conflicts resolving, then we're good to go.

joleb:
@Cepheid-UK conflicts are resolved now :)

PR #390, opened 2021-05-26
reducing bundle size, upgrading packages

joleb:


Cepheid-UK (approved):
Crazy how much these can be compressed with basically no visual difference.

Cepheid-UK:
ok, this one next. Want to rebuild the languages and merge master again? @joleb

joleb:
@Cepheid-UK it's merged now :)

Issue #394, opened 2021-06-04
Minor - widen box for MMR selector


Cepheid-UK:
Doesn't quite fit with some of the translated strings:

![selector2](https://user-images.githubusercontent.com/40397054/120799299-75314080-c536-11eb-8333-b165234be00b.png)

kontrarian:
@Cepheid-UK I'll pick this one up. It looks like a breakpoint issue from a quick inspection.

kontrarian:
It's avatarX on discord btw

Cepheid-UK:
Thanks, and welcome!

kontrarian:
I've created a PR #403 for this with some discussion

Cepheid-UK:
closed by #403

PR #403, opened 2021-07-26
Widen MMR Selector Issue #394

kontrarian:
Changed the column ratios from 3:9 to 4:8 for the Winrates Tab to prevent text from being cutoff when selecting Grandmaster

Before:
Screenshot 2021-07-26 at 12 59 04

After:
Screenshot 2021-07-26 at 13 00 10
Screenshot 2021-07-26 at 13 01 04

### DIFF
--- a/src/components/overal-statistics/tabs/WinratesTab.vue
+++ b/src/components/overal-statistics/tabs/WinratesTab.vue
@@ -1,7 +1,7 @@
 <template>
   <div>
     <v-row>
-      <v-col cols="md-3">
+      <v-col cols="md-4">
         <v-card-text>
           <v-select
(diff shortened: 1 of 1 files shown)


kontrarian:
I'm not familiar with Vuetify but I see it's using a 12 point grid system. Adjusting the column ration from 3:9 to 4:8 fixes the cutoff problem but it looks a bit overly spacious on larger resolutions in my opinion.

I feel a more ideal solution would be to see the selectors on the left to a fixed width and let the statistics table just take the rest of the available space but doing this feels like it's going against the Vuetify grid system which I'm a bit hesitant to do unless there is a really good reason

PR #428, opened 2021-09-19
fixed bug where league icon before season 4 below Master were wrong

Raswanson01:
title sums it up


Cepheid-UK (approved):
checked and its a nice fix, thanks!

Issue #402, opened 2021-07-26
"Europe" region selector is unecessary


Cepheid-UK:
Since we removed the regions, we no longer need to have "Europe" as a selector on the website. (Picture shows on the /Rankings view but it appears in a few other locations too, profile, stats iirc)

I believe the database still keeps data compatible with this region info, so for now we should probably just make the UI change to just remove this selector for now.

From memory, Vuex stores what region you're in, but we can just default everyone to "Europe" (and tbh, maybe we should just rename it to "Worldwide" or something on the enum.

![europeselector](https://user-images.githubusercontent.com/40397054/126967184-1bfe5ef5-cf58-43ec-8d2e-50176c8294e9.PNG)

kontrarian:
Looks like the regions were merged in season 6. What if we just hide the region selector for season > 5, that way players can still see rankings old seasons?

Cepheid-UK:
> Looks like the regions were merged in season 6. What if we just hide the region selector for season > 5, that way players can still see rankings old seasons?

This is a good solution 👍

Cepheid-UK:
closed with #431 ty @Raswanson01

PR #431, opened 2021-09-25
hiding region selector for season 6 and beyond

Raswanson01:
Added a v-if to the gateway-select component to hide for seasons 6 and later


Raswanson01:
#402

Raswanson01:
I didn't commit anything for the profiles or statistics because the page will break for American players. The page will default to America stats, of which there are none, with no selector to get back to Europe. I will look into seeing if I can set the default locale to Europe for season 6+ as well.

Raswanson01:
Actually, this is the case on the rankings page right now, so I'll close this and reopen it once I've figured that out.

Raswanson01:
#432 is the new PR for this

Cepheid-UK:
ty @Raswanson01 this is on the live website now. Sorry for the delay.

PR #434, opened 2021-11-06
fix mobile ui bug on player profile page; closes #421

### LINKED ISSUE #421: Wrong alignment on profile view on mobile screen
Hi I just noticed that the profile page would need to change its flex-direction to `flex-direction: column`.

W3Champions

amacado:
I tried to find a ui solution for #421. I came up with a mixture of grid breakpoints and smaller season badges. It's not 100% ideal, but I think it's already an improvement. I decided to introduce a new property `size` on the `SeasonBadge` component to support further use-cases of this badge.

See screenshots of different screen resolutions:
Bildschirmfoto 2021-11-06 um 02 53 52
Bildschirmfoto 2021-11-06 um 02 53 33
Bildschirmfoto 2021-11-06 um 02 53 15

### DIFF
--- a/src/views/Player.vue
+++ b/src/views/Player.vue
@@ -3,56 +3,50 @@
     <v-row>
       <v-col cols="12">
         <v-card tile>
-          <v-card-title class="justify-space-between">
-            <div
-              style="display: flex; flex-direction: row; align-items: center"
-            >
-              <span>
-                {{ $t("views_player.profile") }} {{ profile.battleTag }}
-              </span>
-              <div v-if="aliasName">({{ aliasName }})</div>
-              <div
-                style="display: flex; flex-direction: row; margin-left: 25px"
-              >
-                <SeasonBadge
-                  v-for="season in seasonsWithoutCurrentOne"
-                  :season="season"
-                  :key="season.id"
-                  :on-click="selectSeason"
-                />
-              </div>
-            </div>
-            <div>
-              <gateway-select @gatewayChanged="gatewayChanged" />
-              <v-menu offset-x v-if="!!seasons && seasons.length > 0">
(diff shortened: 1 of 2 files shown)


src/views/Player.vue
```diff
@@ -3,56 +3,50 @@
...
-                <template v-slot:activator="{ on }">
-                  <v-btn tile v-on="on" class="ma-2 transparent">
+          <v-card-title>
+            <v-row no-gutters>
+              <v-col :align-self="'center'">
+                <span>{{ $t("views_player.profile") }} {{ profile.battleTag }}</span>
+                <span v-if="aliasName" class="ml-1">({{ aliasName }})</span>
+                <span class="mr-2" /> <!-- add some space between name and season badges -->
+                <span v-for="season in seasonsWithoutCurrentOne" :key="season.id" class="ml-1">
+                  <SeasonBadge :season="season" :on-click="selectSeason" :size="'sm'" />
```
Cepheid-UK, line 13: Thanks for the contribution, my only request is to make the `:size` dependent on device size, by hardcoding `sm`, it looks teeny on desktop scale:
image

amacado: The smaller (sm) version was indented for all screen sizes. I think it looks more convienient this way. But sure, I can change that and replace the `size` property with media queries.

amacado: I played a little around and reverted back to the "old" 24px size for all device sizes. It looks decent imo, @Cepheid-UK what do you say?

Bildschirmfoto 2021-11-07 um 01 46 36
Bildschirmfoto 2021-11-07 um 01 47 14
Bildschirmfoto 2021-11-07 um 01 47 27

PR #436, opened 2021-11-07
Update README.md with information for new contributors (development setup)

amacado:
I noticed that the introduction is outdated. I updated and enriched it with my own experiences. 
In a second step, I think we should enhance the workflow so it's not required to modify project files in order to grant yourself admin permissions. I'm thinking of a parameter or environment based approach.

I also fixed a UI bug in "LOADING SCREEN TIPS" (Admin):

<table>
<tr>
<td>Before</td>
<td>After</td>
</tr>
<tr>
<td>

image
</td>
<td>

image
</td>
</tr>
</table>

Happy to hear your feedback!

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,12 +24,12 @@ npm run lint
 
 ### Customize configuration
 
-See [Configuration Reference](https://cli.vuejs.org/config/).
+See [Configuration Reference](https://cli.vuejs.org/config/) for Vue options.
 
 ### Authenticating as Admin
-1) Clone the backend https://github.com/w3champions/w3champions-chat-service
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -24,26 +24,61 @@ npm run lint
 
 ### Customize configuration
 
-See [Configuration Reference](https://cli.vuejs.org/config/).
+See [Configuration Reference](https://cli.vuejs.org/config/) for Vue options.
+
+## Optional: Setup website backend server (and authenticate as admin)
+By default the website you started using `yarn run serve` will connect to the productive
```
Cepheid-UK, line 30: Technically I think it will connect to whatever the last commit to `env.js` was 😃

Issue #407, opened 2021-07-26
Navbar breakpoints need to be adjusted


iamyemeth:
The navbar currently has 3 major breakpoints.

1. Show everything in the navbar
<img width="1208" alt="Screenshot 2021-07-26 at 14 42 50" src="https://user-images.githubusercontent.com/7984606/126999431-939c7c7a-9d4c-4f0e-87d9-be43c42b2c19.png">

2. Show everything but the w3champions logo
<img width="876" alt="Screenshot 2021-07-26 at 14 42 35" src="https://user-images.githubusercontent.com/7984606/126999460-f0bfd656-7bd1-4617-9e9b-05fb8cd3a521.png">

3. Show only the icon links and no logo
<img width="553" alt="Screenshot 2021-07-26 at 14 42 09" src="https://user-images.githubusercontent.com/7984606/126999481-2d286153-6984-4754-905e-575a88002984.png">

The actual breakpoints for these 3 views need to be adjust so there is no cropping of links or navbar scrolling as below. The core idea is that we need to remove the w3champions logo as soon as there is not enough space for all the links and then drop the text links as soon as there is not enough space.

Cepheid-UK:
I'd definitely like to see this cleaner. If anyone has any better ways to breakpoint the navbar I'd welcome it, it takes up a lot of horizontal space.

PR #437, opened 2021-11-07
Redesign main navigation bar & bugfix on /tournaments page

amacado:
- [x] minor code refactorings
- [x] redesign main navigation (for mobile) #407
- [x] code refactoring and component separation

Here are some previous for the updated navbar design:
<table>
<tr>
<td>

image
</td>
<td>

image
</td>
<td>

image

</td>
<td>

image
</td>
</tr>
</table>

And here is a preview for the (partly) fixed bug on the tournaments page:
<table>
<tr>
<td width="30%">

image
image

</td>
<td>

Fixed multiple bugs on [/tournaments page](https://www.w3champions.com/tournaments) which causes the bracked to not show up:

- Duplicate `:keys`
- Invalid access on undefined properties `round.dimensions`
- There is still a bug, described in #439 (but I assume this should be in a new pr)

</td>
</tr>
</table

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -5,8 +5,10 @@
                          v-model="isNavigationDrawerOpen">
       <v-list dense>
         <v-list-item>
-          <v-list-item-content @click="$router.push({ path: '/' })">
-            <brand-logo :is-dark-theme="isDarkTheme"/>
+          <v-list-item-content>
+            <router-link :to="{ name: 'Home'}">
+              <brand-logo :is-dark-theme="isDarkTheme"/>
+            </router-link>
           </v-list-item-content>
           <v-list-item-icon>
             <v-icon class="ml-5" @click="isNavigationDrawerOpen=false">mdi-close</v-icon>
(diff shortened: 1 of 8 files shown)


HooK2000:
I might be missing something here, but vuetify provides the breakpoints as booleans (`$vuetify.breakpoint`), right?

So I'd think a cleaner way than hiding stuff with css would be to actually dismount the components/DOM elements with something like `v-if="$vuetify.breakpoint.lgAndUp"` (desktop) and `v-if="$vuetify.breakpoint.mdAndDown"` (mobile).

amacado:
@HooK2000 you could do that, yeah.. I think it's more a "what do you prefer" decision.. I usually prefer the css solution over the javascript solution for three reasons
- If js is disabled in the browser it still works
- css classes are "more common" and "more known" (personal experience)
- js DOM manipulation probably takes more time than using css (not 100% sure about that)

Is it more readable using `v-if="$vuetify.breakpoint.lgAndUp"`? Maybe. Hard to tell and more a personal preference or?

HooK2000:
@amacado Ok that's fair, I just assumed that this was "the way to go" with vue/vuetify. I'm just a big fan of keeping the DOM clean ;)

amacado:
I haven't found a "best practice" by vuetify, but when they talk about the requirement of visibility changes based on screen sizes they introduce the "css class" approach. See here https://vuetifyjs.com/en/styles/display/#visibility

Issue #451, opened 2022-01-03
When one player from a team is streaming, it shows them all as streaming


Cepheid-UK:
![image](https://user-images.githubusercontent.com/40397054/147982293-e223eb76-5ed8-4321-ad09-4261520023ae.png)

When a player is streaming their game, their teammates show as having active twitch streams, directing to twitch.tv/null.

godrien:
@Cepheid-UK twitch icons are not showing on prod version. https://github.com/w3champions/website/issues/451 also fixes it

Cepheid-UK:
#479 fixes.

PR #454, opened 2022-01-08
styling changes for ranking grid to display cleaner layout for 2v2 an…

debaser19:
…d 4v4 AT


Cepheid-UK:
@debaser19 can you post a screenshot of how this looks different? Just curious before I merge it :D

debaser19:
@Cepheid-UK no worries. keep in mind some of the portraits are not displaying on the dev env as we talked about on discord the other day.

old 2v2
old-2v2

new 2v2
new-2v2

old 4v4
old-4v4

new 4v4
new-4v4

marcoabreu:
Could you show an example with a really long name?

Also, the left and the right cell don't seem to be of the same width. Is that just an illusion?

debaser19:
the longest name i have seen on the rankings has been KODOSFORSAKEN - i'm not sure what the character limit is for the battletags though

i manually added KODOSFORSAKEN to the left cell to test and it was overflowing the portrait, so i bumped the width up a bit

yes, both are the same size, it just looks a bit off as there is a bit of a gap between the player names and the blank Race column

here's a screenshot of the change

image

if i bump the width up higher, it's too wide and wraps to one player per row

marcoabreu:
How about the following:

```
overflow: hidden;
white-space: nowrap;
text-overflow: ellipsis;
```

Rather than tinkering with widths?

Cepheid-UK:
Looks great, can we also have a quick peek at this issue since I think it is affecting the same component, and we might as well put the attention on this to good use:

https://github.com/w3champions/website/issues/451

debaser19:
@marcoabreu i should have some time later tonight to take a look and try that out.

i think the trouble with not using the wrap is that it introduces horizontal scrolling, pushing the win/loss/mmr etc stats out of view. i may be misunderstanding you on this, though

example: 
image

@Cepheid-UK i haven't had a chance to look into the twitch stream logic yet but i'll do some digging and see what i can find

debaser19:
@Cepheid-UK regarding #451, was digging into the twitch logic and i believe this is the issue here

```
 public isTwitchLive(ranking: Ranking): boolean {
    const twitchName = ranking.playersInfo[0].twitchName;
```

looks like you're only checking the player at index 0 if they are live

this explains why ToD is currently streaming, and he has the live twitch badges on the first two listings, but not the third in the below example

image

i've never really picked up javascript, so i'm not immediately sure how to write this, but i believe we need to loop through the list of players on the team to check for truthiness

marcoabreu:
``i think the trouble with not using the wrap is that it introduces horizontal scrolling, pushing the win/loss/mmr etc stats out of view. i may be misunderstanding you on this, though``

That's what 
```
overflow: hidden;
white-space: nowrap;
text-overflow: ellipsis;
```
is for. Note that this part alone is not sufficient. The intention is that we limit the size of the cell to a certain limit. Right now, we allow the cell to grow, wrap and do whatever it wants, resulting in the issues you're observing. Instead, we should clearly limit its boundaries and then use methods to contain the elements within it.

marcoabreu:
You also have to understand that we are talking about three different elements here and I think you're misunderstanding me.

There is the cell: Each cell represents a single player (icon, name, flag and twitch icon)
There is a block: Each block contains multiple cells and represents one entry of a column. 
There is a row: Each row represents one entry and contains multiple blocks that are aligned horizontally with columns.

A cell should not have wrapping enable. It should have very tight boundaries in its size.
A block should wrap so it can place multiple cells into different vertical heights.
A row should not wrap as it otherwise messes up the design. It should only be variable in height, but width should be fixed.

Cepheid-UK:
> i've never really picked up javascript, so i'm not immediately sure how to write this, but i believe we need to loop through the list of players on the team to check for truthiness

Good spot. I can write the fix, or if you want to test your JS chops, you can have a go? Leave the choice up to you.

FHOuH6JXEAMSadX

debaser19:
i was playing around with the twitch logic last night and a bit this morning but wasn't able to get it working. i'd like to see how you fix it though if you can ping me

i made some changes from marco's suggestions, let me know what you guys think.

my changes also pushed out the live game swords icon so i moved that inside the same div as the twitch icon. it also looked like on live the swords icon was at the top of the cell, so i played with the position a bit to center it

old
image


new
image

marcoabreu:
Looks great!

Cepheid-UK:
we happy to merge at this point now @debaser19?

debaser19:
yeah should be good to go

PR #460, opened 2022-01-14
Fix: use https for redirect url

Rotzbua:
Value seems currently not used but better save than sorry 😉


Cepheid-UK:
For clarity, this is only used for dev. When we deploy the app gets env variables from the container. I can merge it in any case tho, just in case someone overlooks this in future.

PR #472, opened 2022-01-23
Fixed missing string for shattered exile in translation file

kangaro0:
Hey,

i noticed that a string for the new shattered exile version was missing in the english translation files, so it would not be shown correctly on the website. See screenshot. 

Issue:
playedmaps

Also in the selector there seems to be some string missing, but it works when i run the project locally. Dont know what happened there. 

Regards,
kangaro0


Cepheid-UK:
Ty for the PR, much tidier!

PR #474, opened 2022-01-27
Update dependencies

kangaro0:
LIBRARY COMMENTS:

**chart.js** 

	Cannot be upgrade to 3.x because of api changes

**@types/moment:**

	Not needed as moment.js provides its own types

**@types/node**

	Added for build scripts 

**sass-loader**

	Throws multiple warnings about sass from vuetify (updated version)
	Have to dig deeper into this, but I think the problem is in another dependency.
        If it's not solvable, I will downgrade again
	https://sass-lang.com/documentation/breaking-changes/slash-div

**babel-runtime:**
	
	Moved to dev dependencies (dependency of vue-markdown)
	Could be replaced by newer @babel/runtime package but owner does not update anymore

**prettier:**

	Updated dependency and adjusted .eslintrc.js and .prettierrc to work properly

**sass-migrator:**

	Added package to migrate sass-code to newer version 
	Replaces division-operator / with math.calc
 

CODE COMMENTS:

**src/views/Player.vue:340**

	Commented out, because it's a read-only property
	Someone has to check whats going on there, seems unnecessary?

**src/components/player/SeasonBadge.vue:34**

	Website throws error at runtime: Season10 Badge not found
	Still loads fine nevertheless, image missing in asset directory

**src/components/ladder/RankingsGrid.vue:**

	Removed public member "sortedRankings" as it causes the component to break
	Saving & manipulating data in @Prop "rankings" works just fine.

**src/components/tournaments/TournamentMatchUpdate.vue:**

	Throws a warning at runtime
	Works fine nevertheless

Other codes changes come from prettier-plugin.

Overview: https://docs.google.com/spreadsheets/d/1TYY8HcFA2S7BP4wbtuZTcPK7Vdgxf9zWHvqXfX6rcsQ


Cepheid-UK:
ty, any idea what this deprecation warning is?
image

I also notice a few errors that look like they might be unsupported stuff now:
image

plus some others that look to be changes to vue?

Seeing as the build went through on azure, I'm thinking something is wrong with the dev environment. `npm install` seemed not to be enough lol.

kangaro0:
Yeah, the first one is an update of the sass language. Issue here is that "/"-operator can mean two things. either it's a division or a seperation operator. So they want to make the language more readable by replacing the old division operator with math.div or calc.

Second one should not be thrown, as i disable the "global-require"-rule of eslint in the config file. 

I'll do some more checking tomorrow

kangaro0:
The error "NodeRequire not defined" comes from the linter. It does not recognize it as a type because it simply does not exist in the web context. The require is used by webpack for including dynamically loaded content during the build process and gets resolved to an ordinary string. I changed the return type of each function returning a "NodeRequire" to "string" and it works like a charm.

Cepheid-UK:
we got some intense deprecation warnings, but this looks good, nice work! 🎉

PR #478, opened 2022-02-04
update mappool translation

debaser19:
Old map pool is still displayed. Updated translations to show current map pool

image


Cepheid-UK:
Ah, that will get overwritten.

When we deploy, the CI runs this script:

https://github.com/w3champions/website/blob/master/scripts/generate-locales.ts

this checks this google doc for the translations and does a deepmerge on the data{} object:

https://docs.google.com/spreadsheets/d/1V5f4zguWDmk9nbnoXSJm9g-ZxImo83NJpSY17EUkzOc

If you want to change translations, we should do it on that sheet. Just request access and ping me (I dont check that google accounts mailbox very often lol)

I think we should delete all the non-english translation blocks in that data object tbh. It's helpful to keep the english translation imo, not least because we currently rely on them for map names, but also so we can easily refer to them in the repo.

PR #506, opened 2022-05-29
fix: map s11 map names W3C-86

godrien:
fixes the issue reported on https://w3champions.atlassian.net/browse/W3C-86


Cepheid-UK:
Use this for translations, but this should be superceded by using the values passed by the API now.

PR #507, opened 2022-06-07
updated other locales to make instructions more clear

gDollaZ:
i left out changes to the other locales previously, including them now.....


Cepheid-UK:
Valuable changes, but they should be done this way:

PR #509, opened 2022-07-05
[W3C-106] Show only first 8 items in carousel. Remove default 500px height

gustav87:
Show only 8 items in carousel instead of all.

Set height to auto instead of default 500px to allow all content to show.

Set a 50 px margin-bottom so that the carousel-controls don't overlap the content.

### DIFF
--- a/src/scss/shared/overrides.scss
+++ b/src/scss/shared/overrides.scss
@@ -197,3 +197,9 @@ header.v-app-bar--is-scrolled + main {
     background-color: var(--w3-bg-glass);
   }
 }
+
+.v-carousel {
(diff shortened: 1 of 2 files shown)


Cepheid-UK:
@gustav87 if you resolve the conflicts on this, I'll merge

cmavelis:
Screen Shot 2022-08-21 at 4 50 31 PM
I'd prefer an alternative to fully uncapped height, since the controls can suddenly jump around pretty far with the announcements that are present. I'll suggest an edit

cmavelis:
merged as part of #535

PR #512, opened 2022-07-21
Fix source maps

gustav87:
Removing yarn.lock and regenerating it with yarn install provides proper source maps with debugging abilities.

image


abeninski:
Very nice

PR #514, opened 2022-07-25
[W3C-143] Query current season API and display maps

gustav87:
image

### DIFF
--- a/src/views/Home.vue
+++ b/src/views/Home.vue
@@ -62,9 +62,6 @@
                 {{ $t("views_home.homebody16") }}
                 <br />
                 <br />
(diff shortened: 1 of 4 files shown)


src/views/Home.vue
```diff
@@ -75,6 +72,20 @@
             </v-row>
           </v-card-text>
         </v-card>
+        <v-card>
+          <v-card-title>{{ $t("views_home.mappooltitle") }}</v-card-title>
+          <div class="seasonMaps" >
```
Cepheid-UK, line 77: this is stylistic, but I really try to avoid using <div>s where ever possible, maybe <ul> might be right here? Semantic tags help SEO.

PR #525, opened 2022-08-05
Update icon logic

BogdanW3:
Rankings and CountryRankings still used the old logic, where TOTAL icons were fallbacks. Now only if there is no selected icon record will a calculated race be used

### DIFF
--- a/src/components/ladder/RankingsGrid.vue
+++ b/src/components/ladder/RankingsGrid.vue
@@ -363,10 +363,7 @@ export default class RankingsGrid extends Vue {
     const playersInfo = ranking.playersInfo;
     if (!playersInfo) return "Random";
     const playerInfo = playersInfo[playerIndex];
-    if (
-      RankingsGrid.hasSelectedIcon(playerInfo) &&
-      playerInfo.selectedRace <= ERaceEnum.UNDEAD
-    ) {
+    if (RankingsGrid.hasSelectedIcon(playerInfo)) {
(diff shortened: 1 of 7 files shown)


src/components/ladder/RankingsGrid.vue
```diff
@@ -363,26 +363,20 @@ export default class RankingsGrid extends Vue {
...
+    if (RankingsGrid.hasSelectedIcon(playerInfo)) {
       return this.$t(`races.${ERaceEnum[playerInfo.selectedRace]}`);
     } else {
       return this.$t(`races.${ERaceEnum[playerInfo.calculatedRace]}`);
     }
   }
 
   private static hasSelectedIcon(playerInfo: PlayerInfo) {
-    if (
+    return (
```
abeninski, line 374: Isn't it better to extract this method in a helper class and use it in both places? Looks identical to me

gustav87:
@BogdanW3 Can you explain again what problem this solves? Is there an example of the issue?

BogdanW3:
> @BogdanW3 Can you explain again what problem this solves? Is there an example of the issue?

This commit was made over two years ago now, the idea was to change up the icon logic to make it more reasonable, but that included some decisions as well. Believe that I have no recollection of the specifics anymore, but something along these lines might still be favourable to pursue, when we figure out what it was.

francislavoie:
I think I'll close this, it's stale and I think these issues have probably been papered over since in one way or another.

PR #531, opened 2022-08-12
Added Dynamic Title for pages

erfanmola:
Hi, I made the page titles dynamic, by default it is set to "W3Champions", But now :
Inside data.ts file, foreach language there is an index called "titles", which holds Titles of the pages in the pair of RouteName:Title
Also for the Profile page which data gets retrevied from server, Page title is set to Player Name & BattleTag


Cepheid-UK:
I appreciate the effort but this will not work with how the system builds the translations.

What is in the repo in that data.ts is only there for development environement.

I recommend reading this doc to understand how localization works on the website:

PR #572, opened 2022-12-26
W3c 195/reenable finished matches

gustav87:
Backend PR:
https://github.com/w3champions/website-backend/pull/219

**Summary of changes**
- Creates a select-season component, and then uses that component in Rankings, CountryRankings and Matches.
- Removes the mmr filter temporarily to not overload the db.
- Removes the refresh interval for finished matches to not overload the db.

I want to push this to Test to see what kind of response times we're getting. It looks OK (not great) on my local machine, but my RAM is maxed out so I'm not sure how representative it is.

**Indexes**
We also need to take a look at the indexes, and perhaps create a new index.

There's an index right now that looks like this:
```
  {
    v: 2,
    key: { EndTime: -1, GameMode: 1, GateWay: 1, Map: 1 },
    name: 'end_games'
  }
```
But that doesn't seem to apply the ESR (Equality, Sort, Range) Rule which is recommended by MongoDB, so I'm not sure how well it performs. I would like to see how that index is performing on Test.

Perhaps one of these (or both) indexes is better suited for the finished matches:
```
  {
    v: 2,
    key: { GateWay: 1, GameMode: 1, Season: 1, EndTime: -1 },
    name: 'end_games2'
  },
  {
    v: 2,
    key: { GateWay: 1, GameMode: 1, Season: 1, Map: 1, EndTime: -1 },
    name: 'end_games3'
  }
```

### DIFF
--- a/src/components/common/SeasonSelect.vue
+++ b/src/components/common/SeasonSelect.vue
@@ -1,13 +1,9 @@
 <template>
   <v-menu offset-x>
     <template v-slot:activator="{ on }">
-      <v-btn
-        tile
-        v-on="on"
-        class="ma-4"
-        style="background-color: transparent"
-      >
-        <h2 class="pa-0">{{ $t("components_common_seasonselect.season") }} {{ selectedSeason.id }}</h2>
+      <v-btn tile v-on="on" :class="{'ma-4': !matchesPage}" style="background-color: transparent">
+        <div v-if="matchesPage"> {{ $t("components_common_seasonselect.season") }} {{ selectedSeason.id }}</div>
+        <h2 v-else>{{ $t("components_common_seasonselect.season") }} {{ selectedSeason.id }}</h2>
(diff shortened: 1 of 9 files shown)


src/components/common/SeasonSelect.vue
```diff
@@ -0,0 +1,57 @@
...
+  @Prop({default: false}) matchesPage?: boolean;
+
+  get seasons() {
+    return this.$store.direct.state.rankings.seasons;
+  }
+
+  get selectedSeason() {
+    return this.matchesPage
+      ? this.$store.direct.state.matches.selectedSeason
+      : this.$store.direct.state.rankings.selectedSeason;
```
stevenaci, line 45: rather do this than make a super class?

gustav87: Hmm, I'm not sure what you mean. Could you elaborate?

PR #606, opened 2023-02-22
Lazy load country flag module

Rotzbua:
## Change

Only load country images if they are used in view.

## Advantage

Reduces initial loading traffic and time. Good for low network speed.

## Disadvantage

Loaded later, delay happen later.

## Neutral

Split into an additional artifact ~200kb:
grafik


gustav87 (approved):

gustav87:
Thanks for the contribution! Lazy loading is indeed a good idea.

Issue #609, opened 2023-02-26
Use `pinia` to replace `vuex`


Rotzbua:
Vue2 is EOL by the end of the year. Migration is difficult. The new default store with vue3 is pinia which also supports vue2. For this reason migration to pinia as store is a good step. Furthermore pinia supports typescript out of the box.

Pinia can be used in parallel to the existing vuex store.

Homepage:
https://pinia.vuejs.org/

Migration:
https://pinia.vuejs.org/cookbook/migration-vuex.html#converting-a-single-module

Cepheid-UK:
I think it would be good to migrate to Vue3 entirely, I believe the main thing holding us back was Vuetify support for it.

PR #610, opened 2023-02-26
Migrate store for twitch to pinia

Rotzbua:
A partial migration as mentioned in #609 .


gustav87:
Hey, great idea to start the migration to pinia. Thank you!

Issue #621, opened 2023-03-06
Remove `moment` dependency


Rotzbua:
The `moment` package is used for time operations.

## Disadvantage

The package not optimized for browsers and has a big bundle size.

## Required changes

* [ ] Should be replaced in project code by the already used `date-fns` which is a lot smaller.
* [ ] Dependency `"chart.js": "^2.9.4"` uses `moment`. Can be solved by upgrading chartjs.

PR #622, opened 2023-03-06
Refactor game duration format

Rotzbua:
## Reasons

`Moment` is not required to format seconds. -> refactored
Locales do not use custom format. -> removed

## Reference

#621


gustav87:
Great stuff man. Something similar could be done for `dateFormats.date` and `dateFormats.dateTime` used in `views/MatchDetail.vue` and `components/matches/MatchesGrid.vue` as well.

Rotzbua:
> Great stuff man. Something similar could be done for `dateFormats.date` and `dateFormats.dateTime` used in `views/MatchDetail.vue` and `components/matches/MatchesGrid.vue` as well.

Already wip ;)

PR #627, opened 2023-03-08
Migrate from `moment` to `date-fns`

Rotzbua:
## Changes

* Use Unicode Tokens https://github.com/date-fns/date-fns/blob/main/docs/unicodeTokens.md
* Create date functions helper
* Use `date-fns` for `chart.js`

## Bugs

* Chart tooltip shows time, should be solved later in the `chart.js` upgrade

## Reference

First part of the solution for #621

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
     "chartjs-plugin-annotation": "^0.5.7",
     "core-js": "^3.27.1",
     "date-fns": "^2.29.3",
+    "date-fns-tz": "^2.0.0",
     "debounce": "^1.2.1",
     "direct-vuex": "^0.12.1",
     "lodash": "^4.17.21",
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
   "dependencies": {
     "@mdi/font": "^6.9.96",
     "chart.js": "^2.9.4",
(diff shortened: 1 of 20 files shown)


gustav87:
The MMR/RP data is on the wrong dates on the player MMR timeline chart because of the UTC conversion to local timezone. Is that fixed when we use a newer version of chartjs, or could we address it now?

Rotzbua:
should be solved with https://github.com/w3champions/website/pull/628/commits/b93f29a984d4597697b64c291375e654faf40699

gustav87:
For some reason, chartjs still puts the wrong dates on the x-axis. The data is on the correct dates now, though, so that's good.
image

Rotzbua:
Thanks for pointing out the problem 👍 

>  For some reason, chartjs still puts the wrong dates on the x-axis

Unfortunately the x-axis is right. The chart starts with `Jan 24, 2023, 11:44pm` but is not labeled because the next tick is just 16 minutes later and for readability reasons the first x-axis-label is `Jan 25`. 🔍 

Of course that is not the intended behavour. The reason is that before only a `string` with the date was provided. Now a `Date` object is used. To get the same result the `Date` has to be set to the start of the day.

gustav87:
Awesome work, man. Merging!

PR #467, opened 2022-01-16
Improve pagespeed

Rotzbua:
Current [pagespeed](https://pagespeed.web.dev/report?url=https%3A%2F%2Fw3champions.com%2F) is not the best so here some improvements.
This PR also includes my other PRs.

PR #634, opened 2023-03-15
Fix & update preconnect / prefetch DNS to improve pagespeed

Rotzbua:
## Change

* `preconnect` (DNS lookup + TLS connection) to the essential domains for a functional website.
* Use only 3 times `preconnect` because more can have negative effect and would exceed the current recommendation.
* Use `dns-prefetch` (DNS lookup) for lower prioritized APIs.
* Both techniques may shorten the delay of the loading process and thus improve the user experience.

## Reference

Fixup of #467 .

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -1,13 +1,17 @@
 <!DOCTYPE html>
 <html lang="en">
   <head>
-
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width,initial-scale=1.0">
     <title>W3Champions</title>
-    <link rel="preconnect" href="<%= BASE_URL %>">
-    <link rel="dns-prefetch" href="<%= BASE_URL %>">
     <link rel="icon" href="<%= BASE_URL %>favicon.ico?v=5">
+
+    <!-- Preconnect TLS and prefetch DNS to improve pagespeed. -->
+    <link rel="preconnect" href="https://w3champions.wc3.tools" crossorigin>
+    <link rel="preconnect" href="https://website-backend.w3champions.com" crossorigin>
+    <link rel="preconnect" href="https://identification-service.w3champions.com" crossorigin>
+    <link rel="dns-prefetch" href="https://api.twitch.tv" crossorigin>
+
     <script type="text/javascript" src="<%= BASE_URL %>env.js?v=4"></script>
   </head>
   <body>


Rotzbua:
> Does the current preconnect + dns-prefetch on the BASE_URL (which is https://w3champions.wc3.tools/prod/site and https://w3champions.wc3.tools/test/site for prod and test, respectively) not work? I didn't find any info on what happens when you pass a path like /prod/site as well as a domain.

`preconnect` and `dns-prefetch` expects a domain.

CanIuse:
tl;dr: Firefox does not support both properly (yet?).
https://caniuse.com/link-rel-dns-prefetch
https://caniuse.com/link-rel-preconnect

Specification by whatwg:
https://html.spec.whatwg.org/#link-type-dns-prefetch
https://html.spec.whatwg.org/#link-type-preconnect

> Is there a before/after comparison of how the website fetches its resources?
> Does it improve pagespeed, and if so how much?

Example:
Request URL: https://website-backend.w3champions.com/api/matches/ongoing?offset=0&gateway=20&pageSize=200&gameMode=1&map=Overall&sort=startTimeDescending

Chrome dev -> Network -> select request -> Timing:
grafik

The delay of 40ms for the connection can be avoided. In the waterfall diagram it is moved forward and processed in parallel. In this case the DNS lookup is just 7us because it was cached by the system.

grafik

With lighthouse / https://pagespeed.web.dev/ it would the effect would be recognizable by a lower LCP (Largest Contentful Paint).

The performance improvement is just for first time visitors and the improvement is by avoiding potential blocking time than in real time saving.

More to read:
https://web.dev/preconnect-and-dns-prefetch/

> Hardcoding the domain like this unfortunately won't work for our test environment https://test.w3champions.com. Even though pagespeed is not important for test, I think it's good to try and keep test and prod as similar as possible.

The `env.js` contains the configuration and is not used while compile time.
I guess it is done to change the app configuration without recompiling but leads to this problem.

gustav87:
env.is is not used at compile time, but `https://w3champions.wc3.tools/prod/site` and `https://w3champions.wc3.tools/test/site` is entered in our build process, as you can see here:

gustav87:
I don't see why the current preload wouldn't work. Is it simply missing the crossorigin attribute?

PR #639, opened 2023-03-18
Use direct import for `lodash` to reduce build size

Rotzbua:
## Changes

* Use direct import for `lodash` to reduce build size
* Add eslint plugin to ensure this for new code

### Difference

Size: -60 KiB
Gzipped: -20 KiB

### Current build

```txt
File                                   Size        Gzipped

dist\js\chunk-vendors.23b7ce38.js      977.84 KiB  281.33 KiB
dist\js\app.24457c62.js                778.11 KiB  296.48 KiB
dist\js\admin-view.d8e430e8.js         368.43 KiB  105.71 KiB
dist\js\840.c7ea0ea6.js                298.85 KiB  216.59 KiB
dist\env.js                            0.98 KiB    0.40 KiB
dist\css\chunk-vendors.01cf3aae.css    746.54 KiB  104.36 KiB
dist\css\app.755083fb.css              26.02 KiB   5.92 KiB
dist\css\admin-view.2e25e0f9.css       5.71 KiB    1.13 KiB
```

### Lodash optimized

```txt
File                                   Size        Gzipped

dist\js\chunk-vendors.f5f0c372.js      916.08 KiB  260.09 KiB
dist\js\app.e5b6a122.js                778.27 KiB  296.64 KiB
dist\js\admin-view.244bfdc7.js         369.80 KiB  106.26 KiB
dist\js\840.c7ea0ea6.js                298.85 KiB  216.59 KiB
dist\env.js                            0.98 KiB    0.40 KiB
dist\css\chunk-vendors.2a195454.css    746.55 KiB  104.36 KiB
dist\css\app.b65f6c24.css              26.10 KiB   5.92 KiB
dist\css\admin-view.2e25e0f9.css       5.71 KiB    1.13 KiB
```

PR #644, opened 2023-03-19
Fix missing lodash functions

Rotzbua:
## Fixes

Hero chart is broken.

## Explanation

Minifying build size does not work properly with `_.chain()`. Some functions are not available at runtime due lazyness. Replaced `chain`.

## Reference 

Fixes previous PR #639

### DIFF
--- a/src/components/overall-statistics/PlayedHeroesChart.vue
+++ b/src/components/overall-statistics/PlayedHeroesChart.vue
@@ -3,10 +3,12 @@
 </template>
 <script lang="ts">
 import Vue from "vue";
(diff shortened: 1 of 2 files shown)


Rotzbua:
@gustav87 lol that timing 😆

gustav87:
And commented simultaneusly. We got to stop this :D

gustav87:
I first tried using lodash/fp instead of creating lots of intermediate variables, but:
1. It didn't work properly and
2. It most likely convolutes the code too much anyway

PR #645, opened 2023-03-21
Refactor services to static service

Rotzbua:
## Note

Contains PR #642

## Reason

* All services are stateless, so no need for service instances.
* Makes migration to `pinia` simpler #609

## Reference

First change with #641


gustav87:
Hey, looking great and a lot simpler than before. Thanks for the hard work. Seems there are some conflicts since the last merge. Could you solve them and repush?

PR #648, opened 2023-03-24
Create lounge chat bans

gustav87:
website-backend PR: https://github.com/w3champions/website-backend/pull/237
matchmaking PR: https://github.com/w3champions/matchmaking-service/pull/294
chat-service PR: https://github.com/w3champions/chat-service/pull/5

### DIFF
--- a/src/components/admin/AdminLoungeMute.vue
+++ b/src/components/admin/AdminLoungeMute.vue
@@ -117,13 +117,13 @@
 <script lang="ts">
 import Vue from "vue";
 import { Component } from "vue-property-decorator";
-import { LoungeMute, LoungeMuteResponse } from "@/store/admin/types";
+import { LoungeMute, LoungeMuteResponse } from "@/store/admin/loungeMute/types";
 import PlayerSearch from "@/components/common/PlayerSearch.vue";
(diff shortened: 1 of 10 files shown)


src/components/admin/AdminLoungeMute.vue
```diff
@@ -0,0 +1,215 @@
...
+      endDate: this.muteEndDateTime,
+    } as LoungeMute;
+
+    await this.$store.direct.dispatch.loungeMute.addLoungeMute(mute);
+    this.loadMutes();
+  }
+
+  get muteEndDateTime(): string {
+    if (!this.endDate) return "";
+    const endDateArr = this.endDate.split("-");
```
abeninski, line 176: This doesn't look good. Why do we do it?

gustav87: To format a date formatted like yyyy-MM-dd to yyyy-mm-ddThh:mm:ssZ. So in other words a UTC time. It's a bit ugly, but it does the job. Reasonably endDate should never be anything but a string formatted like yyyy-MM-dd. But I did a safeguard anyway so if endDate isn't set, we send an empty string to backend, where the request is answered with a proper error code.

PR #668, opened 2023-04-14
Add alt text for screen reader

Rotzbua:


src/views/Home.vue
```diff
@@ -69,7 +69,10 @@
               <v-col cols="12" md="6">
                 <img
                   src="../assets/gettingStarted/startButton_preview_h320.jpg"
-                  style="max-width: 90%; max-height: 320px"
+                  style="width: auto; height: auto; max-width: 90%; max-height: 320px"
+                  width="569"
+                  height="320"
```
gustav87, line 74: Why do we need specify the width and height here?

Rotzbua: Reduces layout shift.
https://pagespeed.web.dev/analysis/https-w3champions-com/bx9s7ew9cp?form_factor=mobile

gustav87: How does setting the logo to 40 px height when it is rendered as 30 px at most reduce layout shift?

Regarding the height and width for the startButton preview. Isn't the height already allocated by the text on the left side of it, and the width by it being part of vue column?

Rotzbua: > How does setting the logo to 40 px height

The logo is not explicitly set to height 40px because this height is overwritten by CSS `  height: auto;
  width: auto;`
The `img` `width` and `height` are used as dimension hint for the browser.

Article: https://web.dev/optimize-cls/


TLDR:

1. styling with attribute
````html
<img src="..." height=40 width=40>
````

2. resource hint and styling with CSS
````html
<img src="..." height=40 width=40>
<style>
img {height: auto; width: auto;}
</style>
````

gustav87: I understand. I just don't see what user benefit it provides. Since these images are contained inside other elements that dictate their size, does setting their explicit size really reduce layout shift? I'm trying to see a difference but I'm unable to. I'm not a fan of more hard coding with numbers without clear references. I'll approve the PR without the layout shift, and only alt texts.

Rotzbua: I am also not a fan of hard-coded numbers but I do not know a programmatically way with vue to take the dimensions from the source file.
Anyway, the the layout shift are not the goal of this PR so now without them.

PR #676, opened 2023-04-23
Remove unnecessary CSS in @font-face

Rotzbua:
Could not find any reason for this code in the documentation:
https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face


gustav87:
Yeah good question. Never seen that, so good catch! :D

gustav87 (approved):

PR #678, opened 2023-04-26
Migrate font `inter` from static asset to package dependency

Rotzbua:
## Fix

Missing used font weights: 500, 700

## Solution

Use variable weight font. Smaller than loading 4 different font weights. (~60kb vs ~30kb)


gustav87:
I've only pulled this down and looked at it quick, but it seems to me that the `bolder` font weight is displayed as `bold`. I'm looking at the text displayed in the carousel on the front page.

Rotzbua:
Do you mean the text "Settings -> Switch To Live"?
This is actually font weight 700. This font weight is not explicitly supplied. The browser downsizes the font weight from the available 900. This leads to an alias effect and the font looks a bit bigger and different than the "true" 700 font.

gustav87:
I mean this text here.
image

Rotzbua:
Ok. I am on the test env.

Bolder is a relative value. The normal text is 400 and the bolder weight is 700. So the browser downsizes it from the available 900.
https://developer.mozilla.org/en-US/docs/web/CSS/font-weight#meaning_of_relative_weights

variable_test

gustav87:
Now I understand what you mean by that it fixes the missing font weights 500 and 700. That is indeed a good change. Just one thing. `src/scss/fonts/inter.scss` looks a little messy, with quadruple `/` in front of some weights and double in front of some. Could you clean that up?

src/scss/fonts/inter.scss
```diff
@@ -1,21 +1,20 @@
...
+https://fontsource.org/fonts/inter
+*/
+////@import "@fontsource/inter/100.css"; /* Thin */
+////@import "@fontsource/inter/200.css"; /* ExtraLight */
+////@import "@fontsource/inter/300.css"; /* Light */
+//@import "@fontsource/inter/400.css"; /* Regular */
+//@import "@fontsource/inter/500.css"; /* Medium */
+////@import "@fontsource/inter/600.css"; /* SemiBold */
+//@import "@fontsource/inter/700.css"; /* Bold */
+////@import "@fontsource/inter/800.css"; /* ExtraBold */
```
gustav87, line 17:
```suggestion
//@import "@fontsource/inter/100.css"; /* Thin */
//@import "@fontsource/inter/200.css"; /* ExtraLight */
//@import "@fontsource/inter/300.css"; /* Light */
//@import "@fontsource/inter/400.css"; /* Regular */
//@import "@fontsource/inter/500.css"; /* Medium */
//@import "@fontsource/inter/600.css"; /* SemiBold */
//@import "@fontsource/inter/700.css"; /* Bold */
//@import "@fontsource/inter/800.css"; /* ExtraBold */
```

gustav87 (approved):

PR #691, opened 2023-06-10
W3C-206/Display description about game modes in more places

mattixpet:
Display description about game modes being normalized also when affected game modes are selected.

That is the 2v2 is counted twice, 4v4 4 times, etc.

### DIFF
--- a/src/components/overall-statistics/tabs/PlayerActivityTab.vue
+++ b/src/components/overall-statistics/tabs/PlayerActivityTab.vue
@@ -19,7 +19,7 @@
               outlined
             />
           </v-card-text>
-          <div v-if="isAllMode">
+          <div v-if="displayDetailedModeDescription">
             {{
               $t(
                 "components_overall-statistics_tabs_playeractivitytab.gamemodedesc1"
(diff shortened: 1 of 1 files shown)


gustav87:
You should pull game modes from backend instead of hardcoding them in the frontend.

mattixpet:
Thanks for the reply, didn't know that was possible, where is the documentation or code for that? However, the EGameMode.UNDEFINED is already hardcoded there, so I don't see how this is much worse than the current code.

I was trying to be fancy by not showing the text for unaffected game modes like 1v1, it's possible to solve this issue in a simpler way by just showing the text always, I think that's better than letting it disappear like is currently the situation. Then there's no hardcoding, would that fly better?

Also, it's pretty arbitrary which game modes are not included (okay it's not at all, but they don't seem to be tagged with any info far as I can tell, they are only enums, so this needs to be hardcoded anyway). If there was some way to see if a game mode was 1v1 or multiplayer that could be used instead of hardcoding (e.g. string search (e.g. "1V1_") or looking up a property of an object).

PR #728, opened 2024-01-25
feat: add engine requirement

Rotzbua:
To avoid wrong node version.


gustav87:
Hi, I added a .nvmrc file to the repo to indicate the proper Node version to use during development. The production Node version is listed in the Dockerfile. Is there another reason to add it to the package.json file except informing developers?

Issue #577, opened 2023-01-13
Switch to `vite`


Rotzbua:
`vue-cli` is in maintenance mode. 
https://github.com/vuejs/vue-cli#%EF%B8%8F-status

The new default build system is `vite`. It is faster and offers a simpler build process.

I have a working first draft. PR coming soon.

update:
It seems that your app have with some components an event emit and scope problem. Not sure where this comes from.

gustav87:
Hey Rotzbua, do you have an example of an event emit and scope problem? I'm not exactly sure what you mean.

Rotzbua:
Click on profile button and nothing happens. The browser console logs following error:
```txt
[Vue warn]: Error in v-on handler: "TypeError: this is null"

found in

---> <VBtn> at website/src/App.vue
       <VAppBar> at website/src/App.vue
         <VApp> at website/src/App.vue
           <App> at website/src/App.vue


TypeError: this is null
    loginOrGoToProfile App.vue:220
    VueJS 4
    click VBtn.ts:163
    VueJS 63
    <anonymous> main.ts:60
    VueJS 7
    click VBtn.ts:163
    VueJS 63
    <anonym> main.ts:60

gustav87:
I checked this out and `this` is null in that context for some reason. Making `loginorGoToprofile` an arrow function "solves" it, but there's evidently some larger issue at play. I played around with it for a while but it just gave me a headache! Is it better to switch to Vite at the same time as Vue 3? Perhaps using Vite and Vue2 is difficult.

Rotzbua:
I agree with you that the change should be later. Maybe a dependency is the reason for the problem or some bug in the code.
I would not pursue this further at this time, but leave the issue open to gather more information.

gustav87:
Completed in https://github.com/w3champions/website/pull/734.

PR #734, opened 2024-03-23
Add Vite to project

gustav87:


gustav87:
@Rotzbua Hi, I continued your work with adding Vite to the project. I'm not getting the error that was mentioned in https://github.com/w3champions/website/issues/577, nor can I remember how that happened.

It looks pretty good locally right now, and when building locally. Do you know anything else we need to add here?

gustav87:
We remove the Vuetify dateTimePicker, because it causes issues. The app can't run when I do `yarn dev`, because it cannot resolve the path of the DateTimePicker. It's only used when creating tournaments in the admin panel, which is a rare occurrence for now. We will replace it with something else.

PR #742, opened 2024-03-26
Upgrade to vue 3, vuetify 3. only for reference (do not merge)

gustav87:


gustav87:
@Rotzbua I started migrating to Vue 3 and Vuetify 3. It's been messy and frustrating, but it actually runs now, at least. I've stripped away themes, cookies and some locales for now. It's difficult to work incrementally, because I needed to upgrade Vuetify, Router, Tiptap and fix the initialization of i18n (translations), pinia and probably more that I can't remember at the moment.

Shoot me a message if you want to help out. This is the state of it at the moment. As you can see, it renders and it fetches data properly, so that's pretty cool.
image

gustav87:
It's very messy right now, and I've added another material design package, just because I couldn't figure out how to use the existing one. Icons didn't render. I'm doing something wrong of course, but right now I just want to see if we can get it into a okayish state without too much work.

gustav87:
@Rotzbua Thanks for your feedback. I had big problems getting the themes to work, and realized it didn't work properly until I moved App.vue to the composition API. I added a new lib called `vue-facing-decorator` to continue using class style syntax, but for some reason, I can't get theming and classes to work. It works without problem after I moved to the composition API.

Perhaps it is wise to move to the Vue Composition API in Vue 2 before attempting the migration. What do you think? We can leave this PR here for reference.

PR #743, opened 2024-03-27
Composition API for App.vue

gustav87:
The reason for this migration is to make the migration to Vue 3 easier later. It was quite troublesome to work with classes in Vue 3 (I had issues getting the theme to work properly), and official stance from the Vue team is to either use the composition api or options api (https://vuejs.org/guide/extras/composition-api-faq.html#relationship-with-class-api).

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,7 +63,7 @@
     "@types/prettier": "^2.7.3",
     "@typescript-eslint/eslint-plugin": "^5.62.0",
     "@typescript-eslint/parser": "^5.62.0",
-    "@vitejs/plugin-legacy": "^5.3.2",
+    "@vitejs/plugin-legacy": "^4.1.1",
     "@vitejs/plugin-vue2": "^2.3.1",
     "@vue/eslint-config-prettier": "^7.0.0",
     "@vue/eslint-config-typescript": "^11.0.2",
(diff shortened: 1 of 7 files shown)


gustav87:
@Rotzbua I made an attempt to migrate App.vue to the composition API as a starting point. Probably not the easiest module to migrate, but serves as a good test to see if it's a reasonable effort.

I'm new to the composition API so I don't know what I'm doing. I managed to get everything working except auth. You can auth, but `authCode` is not properly populated after you log in, so you can't go to your profile. Do you have any idea..? :D

gustav87:
@Rotzbua Hi, I fixed the auth issue and so the composition api for App.vue seems to work good now.

The reason for this migration is to make the migration to Vue 3 easier later. It was quite troublesome to work with classes in Vue 3 (I had issues getting the theme to work properly), and official stance from the Vue team is to either use the composition api or options api (https://vuejs.org/guide/extras/composition-api-faq.html#relationship-with-class-api).

I also had to downgrade Vite from v5 to v4, because there was a problem with hot reloading. After a hot reload, I always got a reference error.


I think probably my way of structuring the code in App.vue is bad, and could be done better, but functionality-wise, it seems to run like before.

PR #794, opened 2024-05-26
Add SeasonSelect component for matches view

healfy:
### Summary of changes

Creates a select-season component, and then uses that component for the Matches query.

### DIFF
--- a/src/components/common/SeasonSelect.vue
+++ b/src/components/common/SeasonSelect.vue
@@ -33,7 +33,6 @@ import { useMatchStore } from "@/store/match/store";
 
 export default defineComponent({
   name: "SeasonSelect",
-  props: {},
   setup: (_, context) => {
     const rankingsStore = useRankingStore();
     const matchStore = useMatchStore();
--- a/src/components/common/SeasonSelect.vue
+++ b/src/components/common/SeasonSelect.vue
@@ -1,0 +1,56 @@
+<template>
+  <v-menu offset-x>
+    <template v-slot:activator="{ on }">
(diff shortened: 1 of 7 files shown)


src/components/common/SeasonSelect.vue
```diff
@@ -0,0 +1,56 @@
...
+
+<script lang="ts">
+import { computed, ComputedRef, defineComponent } from "vue";
+import { Season } from "@/store/ranking/types";
+import { useRankingStore } from "@/store/ranking/store";
+import { useMatchStore } from "@/store/match/store";
+
+export default defineComponent({
+  name: "SeasonSelect",
+  props: {},
```
gustav87, line 36: Delete empty props property.

src/store/types.ts
```diff
@@ -216,3 +216,7 @@ export enum EComputer {
   NORMAL = 1,
   INSANE = 2,
 }
+
+export type Season = {
+  id: number;
+};
```
gustav87, line 222:
```suggestion
```

This type already exists in `src\store\ranking\types.ts`

src/components/common/SeasonSelect.vue
gustav87, line 36:
```suggestion
```
Empty props not needed. The linter complains about that when upgrading to Vue 3, so might as well delete it now.

src/store/types.ts
healfy: fixed

src/components/common/SeasonSelect.vue
healfy: fixed

PR #804, opened 2024-06-01
Auth headers admins API

healfy:
put token to headers rest of admins API


src/helpers/date-functions.ts
```diff
@@ -41,3 +41,32 @@ export const dateToCurrentTimeDate = (endDateString: string): string => {
   now.setFullYear(endDate.getFullYear());
   return now.toISOString();
 };
+
```
gustav87, line 44: Should this not be in some other place? We put date related functions here.

healfy: I misread date with data:d 
Sure, I will move it to another place

gustav87 (approved):

PR #810, opened 2024-06-11
Resubmit translation revision

jzy-chitong56:
It seems that some variables are illegal, and I'm not sure if they were intentional. I understand them as input errors and have made some corrections


gustav87:
This file is automatically generated from a google excel sheet containing the translations. We do not modify this file manually.

gustav87:
Read this for further info: https://github.com/w3champions/website/pull/531

Thank you for your contribution though! Appreciate it.

PR #824, opened 2024-08-21
Matches - use internal page state when polling for matches to prevent page resetting

sivertsenstian:
The refresh interval was setting the page to 1, and the page was also updated if undefined when getting matches.
Now always uses the internal state directly, and the page has to be updated using the setter (e.g when changing filters)

Also add retry logic if the last page of ongoing matches is selected, but the number of matches are reduced  on update, removing the current page.

### DIFF
--- a/src/store/match/store.ts
+++ b/src/store/match/store.ts
@@ -34,6 +34,14 @@ export const useMatchStore = defineStore("match", {
           this.mmr,
           this.sort
         );
+
+        // Handle edge case when loading ongoing matches, if the number of matches are reduced
+        // so that the current page no longer exists, we decrement the current page by 1 and retry.
+        if (!response.matches.length && this.page > 1) {
+          this.SET_PAGE(this.page - 1);
+          await this.loadMatches();
+          return;
+        }
       } else {
         response = await MatchService.retrieveMatches(
           this.page - 1,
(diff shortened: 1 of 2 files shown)


src/views/Matches.vue
```diff
@@ -158,8 +146,8 @@ export default defineComponent({
     onMounted(async (): Promise<void> => {
       await rankingsStore.retrieveSeasons();
       rankingsStore.setSeason(rankingsStore.seasons[0]);
-      matchStore.setSeason(rankingsStore.seasons[0]);
-      getMatches(1);
+      await matchStore.setSeason(rankingsStore.seasons[0]);
```
gustav87, line 149: Looks like this call already loads matches, so the line below would load them twice.

src/store/match/store.ts
```diff
@@ -22,8 +22,7 @@ export const useMatchStore = defineStore("match", {
     selectedSeason: {} as Season,
   }),
   actions: {
-    async loadMatches(page?: number) {
```
gustav87, line 25: This gets called from more places than Matches.vue, for example PlayerMatchesTab.

gustav87 (changes requested):

src/views/Matches.vue
sivertsenstian: Ah true, I just added the missing await as I was getting a warning. Will remove the redundant call

src/store/match/store.ts
sivertsenstian: I tried checking other uses for this, but couldn't find any. The one you mention seems to be using a similar function, but in the `playerStore` and not `matchStore`

gustav87 (approved):

PR #825, opened 2024-08-22
Minor improvements to player profile page

sivertsenstian:
- Re-align the league banner and center stats by mode (as well as using the entire column space available)
- Use icons instead of using `W/L` for recent performance

Current:
image

New:
image
image
image

### DIFF
--- a/src/components/player/ModeStatsGrid.vue
+++ b/src/components/player/ModeStatsGrid.vue
@@ -11,7 +11,7 @@
     <template v-slot:body="{ items }">
       <tbody>
         <tr v-for="item in items" :key="item.id">
-          <td class="cell mode">
+          <td class="cell d-flex justify-center align-center">
             <span>{{ $t("gameModes." + EGameMode[item.gameMode]) }}</span>
             <race-icon style="display: inline; padding-left: 10px" :race="item.race" />
           </td>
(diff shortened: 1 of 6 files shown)


src/views/Player.vue
```diff
@@ -352,6 +315,13 @@ export default defineComponent({
 </script>
 
 <style lang="scss" scoped>
+.red {
```
gustav87, line 318: Where is this 'red' class used?

sivertsenstian: https://github.com/w3champions/website/blob/e3cf691a3b2a9dd2f3e58ea6b1c06044995dd13c/src/views/Player.vue#L50

I could only find it used once in `Player.vue` so I moved it from global to scoped for the component. It was messing up the red `L` icons.

src/components/player/ModeStatsGrid.vue
```diff
@@ -181,4 +172,14 @@ export default defineComponent({
 .tooltip-inner {
   white-space: pre-line;
 }
+
+.cell {
+  white-space: nowrap;
+
+  &.mode {
```
gustav87, line 179: Instead of creating this class, you can use vuetify html classes: `d-flex justify-center align-center`

gustav87 (approved):

PR #831, opened 2024-09-15
style: fix eslint checking

gweesin:
fixes for these:

1. TypeScript definition `any`
2. image should contains `alt` attribute
3. unused variables


src/components/player/PlayerHeroStatistics.vue
```diff
@@ -103,7 +103,7 @@ export default defineComponent({
         heroStatsData.push(rowObject);
       }
 
-      let tableData: any[] = [];
+      let tableData: never[] = [];
```
gustav87, line 106: Can you explain the rationale for using never here?

src/components/player/PlayerHeroStatistics.vue
```diff
@@ -126,7 +126,7 @@ export default defineComponent({
       });
     }
 
-    function populateDataForTable(tableData: any[], heroStatsData: any[]) {
+    function populateDataForTable(tableData: never[], heroStatsData: never[]) {
```
gustav87, line 129: And here.

src/components/tournaments/Tournament.vue
```diff
@@ -19,6 +19,7 @@ import { Map } from "@/store/admin/mapsManagement/types";
 import { useTournamentsStore } from "@/store/tournaments/store";
 
 export default defineComponent({
+  // eslint-disable-next-line vue/multi-word-component-names
```
gustav87, line 22: We should fix it instead. Could name it TournamentView in lack of other ideas.

src/views/MatchDetail.vue
```diff
@@ -345,6 +345,7 @@ export default defineComponent({
       isCompleteGame,
       scoresOfWinners,
       scoresOfLosers,
+      ffaLoser1,
```
gustav87, line 348: Is this used?

src/views/Matches.vue
```diff
@@ -133,6 +133,7 @@ export default defineComponent({
       overallStatsStore.loadMapsPerSeason();
     }
 
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
gustav87, line 136: Should fix instead of ignore.

src/views/Tournaments.vue
```diff
@@ -58,6 +58,7 @@ export default defineComponent({
     return {
       upcomingTournaments,
       tournaments,
+      pastTournaments,
```
gustav87, line 61: Is it used?

PR #836, opened 2024-09-21
improved padding on match details view card

elevchyt:
Simple UI improvement I did to play around the project and learn its structure for future issues.

before
after

### DIFF
--- a/src/views/MatchDetail.vue
+++ b/src/views/MatchDetail.vue
@@ -3,7 +3,7 @@
     <div v-if="isJubileeGame" class="jubilee"></div>
     <v-row v-if="!loading">
       <v-col cols="12">
-        <v-card tile class="pa-6">
+        <v-card tile class="pb-5">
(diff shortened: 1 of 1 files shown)


gustav87:
Hi, thanks for the contribution! I agree it needs some padding, but pa-8 seems excessive.

elevchyt:
> Hi, thanks for the contribution! I agree it needs some padding, but pa-8 seems excessive.

No problem, just pushed a new commit that sets it to 6.

gustav87:
image
Ideally these should be the same, I think

gustav87:
So you may only need to add padding to the bottom

elevchyt:
I see, thought the names should have some more padding since their container has some padding by itself.
Added padding bottom of 5 so that it's equal to the top side padding.

gustav87 (approved):

gustav87:
Nice, thank you for the contribution!

Issue #845, opened 2024-10-17
Questions about Games Per Day stats


francislavoie:
Obviously Direct Strike has had a huge surge in popularity https://w3champions.com/OverallStatistics/

This brings up these questions:

- Are DS games counted multiple times due to being 3v3 like the sidebar mentions 2v2/4v4/FFA are? I think it seems to be by looking at `https://website-backend.w3champions.com/api/w3c-stats/games-per-day` output, the numbers are exactly 1/3 of what appears on the graph. Maybe the text on the left can be revised to explain that custom games are also normalized? Might be interesting to add a checkbox/switch to disable normalizing (should be possible since normalizing is done clientside it seems)?

- How is "All" calculated? Is it counted server-side? Is it using a specific list of modes to total up? It's impossible that it's including Direct Strike games because DS has even surpassed "All" now :joy: unless it only counts pre-normalized numbers?

- Is there a reason the graph only shows data from 2 days prior? It seems to have data for even current. I get that it makes sense to skip today's numbers because it would mean the graph would have a hard dip until the day is over, but shouldn't it be safe to show 1 day prior?

PR #846, opened 2024-10-18
Clean up games per day graph

francislavoie:
Closes https://github.com/w3champions/website/issues/845

- Fixes the date calculation for the graph; previously it would use the local date when the data was UTC midnight, so specifically for anyone with a timezone west of UTC, it would show the previous date rather than the date of the data. 
- Added a "Normalized" switch which allows turning off the normalization multiplier per game mode.
- Added the missing `GM_WARHAMMER_1ON1` entry to `en.ts` (oversight from https://github.com/w3champions/website/pull/843), also fixed `Platin`->`Platinum` for english
- Wired up translation for the "Games per Day" title and "Normalized" switch
  - I used google translate to do my best to fill them in, but surely they'll need a pass from whoever speaks those languages.
  - I skipped the SR language though because it's... horribly broken. If you take a look through it, it's obvious that the translation strings are mismatched to their keys, like they're all off by one. For example:
    ```
    winlossdesc: "MMR",
    mmr: "MMR",
    mmrdesc: "RP",
    rp: "Nivo",
    ```
    All the values are shifted up by one key. I don't speak the language so I don't feel confident to try to fix it, nor fill many of the missing keys.
- Slightly adjusted the layout:
  - Remove the `v-card-text` around the row which added padding, and to be consistent with the other graphs below
  - Add `hide-details` to the `v-select` which removes some dead space below it

Before:
image

After:
image

### DIFF
--- a/src/components/overall-statistics/ActivityPerDayChart.vue
+++ b/src/components/overall-statistics/ActivityPerDayChart.vue
@@ -10,6 +10,7 @@ import LineChart, { getBackgroundColor } from "@/components/overall-statistics/L
 import { ChartData, ScriptableContext } from "chart.js";
 import { EGameMode } from "@/store/types";
 import { parseJSON } from "date-fns";
+import { utcToZonedTime } from "date-fns-tz";
 
 export default defineComponent({
   name: "ActivityPerDayChart",
@@ -23,6 +24,10 @@ export default defineComponent({
     },
     selectedGameMode: {
       type: Number as PropType<EGameMode>,
+      required: true,
+    },
+    normalized: {
+      type: Boolean,
       required: true,
     },
   },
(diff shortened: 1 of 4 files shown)


francislavoie:
@gustav87 I'd love a review if you have time :relaxed: 

I was also thinking we should probably ignore the "All" line from the backend and recalculate it using the normalization so that it actually totals the numbers that appear on the graph. I can implement that as well if you agree.

gustav87:
Hey! This looks great. Regarding the translations, the translations from the files are not used in production. 

https://github.com/w3champions/website?tab=readme-ov-file#working-with-localization
https://github.com/w3champions/website/blob/master/scripts/generate-locales.ts

We add translations by editing the google sheet. The translation files will be overwritten in production upon deploy. Every now and then, we run the generate-locales.ts script locally and push the changes to update the translations for local development.

Are you a member of the w3c discord? I can give you the dev role so you can access the internal development channels.

gustav87 (approved):

francislavoie:
I see, will you add the couple new translations to the doc?

Is there possibly a bug with the generator script, or has it just not been run in a while (regarding SR)

I am on Discord, but I'm not too active: `ruhigengeist`

gustav87:
I would suspect the translations are shifted by mistake in the google sheet. I gave you the staff role in the Discord so you have access to the internal channels, if you want to discuss something.

PR #851, opened 2024-10-26
GamesPerDay: Recalculate "All" as normalized

francislavoie:
This normalizes the "All" line on the games per day graph, by recalculating the data from all the other game modes, ignoring the value from the backend, when Normalized is switched on. This makes sure that "All" isn't smaller than popular modes like Direct Strike that have a multiplier.

Also increased the `pointHitRadius` on all the line graphs so that hovering on specific datapoints is easier.

I also cleaned up some type errors I noticed from clicking around on the various graph components.

Before:

image

After:

image

### DIFF
--- a/src/components/overall-statistics/ActivityPerDayChart.vue
+++ b/src/components/overall-statistics/ActivityPerDayChart.vue
@@ -43,7 +43,10 @@ export default defineComponent({
     );
 
     // Get the list of dates for the x-axis
-    const gameDayDates = computed(() => allSet.value.gameDays.map((g) => utcToZonedTime(parseJSON(g.date), "UTC")));
+    const gameDayDates = computed(() => allSet.value.gameDays
+      .slice(0, allSet.value.gameDays.length - 1) // Drop today's data (last entry)
+      .map((g) => utcToZonedTime(parseJSON(g.date), "UTC"))
+    );
 
     // Recompute the "All" set using all the other game modes,
     // using each mode's normalizing multiplier (the backend doesn't normalize)
@@ -94,6 +97,7 @@ export default defineComponent({
             borderColor: mapColor(c.gameMode),
             borderWidth: 1.5,
             tension: 0.4, // Smooth line.
+            pointHitRadius: 6,
           } satisfies ChartDataset<"line">)),
       };
     });
--- a/src/components/overall-statistics/ActivityPerDayChart.vue
+++ b/src/components/overall-statistics/ActivityPerDayChart.vue
@@ -2,12 +2,12 @@
   <line-chart :chart-data="gameHourChartData" />
 </template>
 <script lang="ts">
-import { computed, ComputedRef, defineComponent, PropType, ref } from "vue";
+import { computed, defineComponent, PropType } from "vue";
 import { useI18n } from "vue-i18n-bridge";
 import { activeGameModes } from "@/mixins/GameModesMixin";
-import { GameDayPerMode } from "@/store/overallStats/types";
+import { GameDay, GameDayPerMode } from "@/store/overallStats/types";
 import LineChart, { getBackgroundColor } from "@/components/overall-statistics/LineChart.vue";
-import { ChartData, ScriptableContext } from "chart.js";
+import { ChartData, ChartDataset, Point } from "chart.js";
 import { EGameMode } from "@/store/types";
 import { parseJSON } from "date-fns";
 import { utcToZonedTime } from "date-fns-tz";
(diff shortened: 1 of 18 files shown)


src/components/player/PlayerHeroWinRate.vue
```diff
@@ -105,7 +103,7 @@ export default defineComponent({
       { text: "vs. Random", value: ERaceEnum.RANDOM },
     ];
 
-    const headersWithoutImageAndName = headers.slice(2);
+    const headersWithoutImageAndName = headers.slice(2) as { text: string; value: ERaceEnum }[];
```
gustav87, line 106: What is the benefit of this `as`?

src/components/overall-statistics/AmountPerDayChart.vue
```diff
@@ -23,25 +24,24 @@ export default defineComponent({
...
-            fill: true,
-            backgroundColor: "rgba(54, 162, 235, 0.2)",
-            borderColor: "rgba(54, 162, 235, 1)",
-            borderWidth: 1.5,
-            tension: 0.4, // Smooth line.
-          },
-        ],
-      };
-    });
+    const chartData = computed<ChartData<"line">>(() => ({
```
gustav87, line 30: Why use an implicit return statement here? If it's a oneliner function, it's nice, but doesn't improve readability if the function spans multiple lines, in my opinion.

src/components/player/PlayerHeroWinRate.vue
francislavoie: TS thinks it's `{ text: string; value: ERaceEnum } | { text: string; value: string }` without this, because that's effectively what `headers` has. Then the template above at `<td v-on="on" v-html="item[header.value]"></td>` would complain cause it doesn't know how to use `string` as an index into an array (strings can index into an array).

```
Element implicitly has an 'any' type because expression of type 'string | ERaceEnum' can't be used to index type 'PlayerHeroWinRateForStatisticsTab'.
  No index signature with a parameter of type 'string' was found on type 'PlayerHeroWinRateForStatisticsTab'.ts-plugin(7053)
```

src/components/overall-statistics/AmountPerDayChart.vue
francislavoie: Removes a level of indentation, less to think about overall IMO. But fair enough, I can revert that if you prefer.

gustav87: It's a minor gripe, and you probably have more professional experience than me, so if you tell me it's stylistically kosher, I trust you.

src/components/overall-statistics/ActivityPerDayChart.vue
```diff
@@ -34,44 +34,71 @@ export default defineComponent({
...
-              borderWidth: 1.5,
-              tension: 0.4, // Smooth line.
-            };
-          }),
+          // Use the calculated total for "All" (UNDEFINED) if normalized is on
+          .map((c) => c.gameMode == EGameMode.UNDEFINED && props.normalized ? normalizedAllSet.value : c)
+          // Filter out inactive gamemodes, but show total games (UNDEFINED)
+          .filter((c) => activeGameModeIds.includes(c.gameMode) || c.gameMode === EGameMode.UNDEFINED)
+          // Only show the mode that user selected (but "All" shows everything)
+          .filter((c) =>
```
gustav87, line 78:
```suggestion
          .filter((c) =>
```

src/components/overall-statistics/AmountPerDayChart.vue
```diff
@@ -2,12 +2,13 @@
   <line-chart :chart-data="chartData" />
 </template>
 <script lang="ts">
-import { computed, ComputedRef, defineComponent, ref } from "vue";
+import { computed, defineComponent, ref } from "vue";
 import { useI18n } from "vue-i18n-bridge";
 import { GameDay } from "@/store/overallStats/types";
 import LineChart from "@/components/overall-statistics/LineChart.vue";
-import { ChartData } from "chart.js";
+import { ChartData, ChartDataset } from "chart.js";
```
gustav87, line 9:
```suggestion
import { ChartData } from "chart.js";
```

src/components/overall-statistics/LineChart.vue
```diff
@@ -20,9 +20,12 @@ import {
   Legend,
   ChartArea,
   ScriptableContext,
+  ChartData,
+  Point,
```
gustav87, line 24:
```suggestion
```

gustav87:
There's some linting to do for unused variables. Approved when fixed.

gustav87 (approved):

PR #852, opened 2024-10-26
PlayerStats: Find most played mode, redetermine on season switch

francislavoie:
When looking at a player's stats page, the MMR timeline graph doesn't correctly choose the player's favourite mode, it always defaults to 1v1. Lately, some players (e.g. Grubby) have Direct Strike as their most played instead, so it's nice to default to that right away.

Also, when switching seasons, we should recalculate their favourite mode & race for that season. To do that, we need to await fetching the stats for the current season before calling `initMmrRpTimeline` which uses it as a dependency.

Also, the page should load significantly faster now, we can parallelize loading a bunch of the data with `Promise.all()` instead of chained `await`

### DIFF
--- a/src/views/Player.vue
+++ b/src/views/Player.vue
@@ -280,11 +280,9 @@ export default defineComponent({
     }
 
     async function initMmrRpTimeline() {
-      const selectedSeason = playerStore.selectedSeason?.id;
       let maxMode = EGameMode.GM_1ON1;
       let maxModeGames = 0;
       playerStore.gameModeStats.forEach((m) => {
-        if (m.season !== selectedSeason) return;
         if (m.games > maxModeGames) {
           maxModeGames = m.games;
           maxMode = m.gameMode;
@@ -295,7 +293,6 @@ export default defineComponent({
       let maxRace = ERaceEnum.HUMAN;
       let maxRaceGames = 0;
       playerStore.raceStats.forEach((r) => {
-        if (r.season !== selectedSeason) return;
         if (r.games > maxRaceGames) {
           maxRaceGames = r.games;
           maxRace = r.race;
--- a/src/views/Player.vue
+++ b/src/views/Player.vue
@@ -224,15 +224,18 @@ export default defineComponent({
       );
     }
 
-    function selectSeason(season: Season): void {
+    async function selectSeason(season: Season): Promise<void> {
       playerStore.SET_SELECTED_SEASON(season);
-      playerStore.loadGameModeStats({});
-      playerStore.loadRaceStats();
-      playerStore.loadMatches(1);
-      playerStore.loadPlayerStatsRaceVersusRaceOnMap(battleTag.value);
-      playerStore.loadPlayerStatsHeroVersusRaceOnMap(battleTag.value);
-      playerStore.loadPlayerMmrRpTimeline();
-      playerStore.loadPlayerGameLengths();
+      await Promise.all([
+        playerStore.loadGameModeStats({}),
+        playerStore.loadRaceStats(),
+        playerStore.loadMatches(1),
+        playerStore.loadPlayerStatsRaceVersusRaceOnMap(battleTag.value),
+        playerStore.loadPlayerStatsHeroVersusRaceOnMap(battleTag.value),
+        playerStore.loadPlayerGameLengths(),
+      ]);
+      // This requires loadGameModeStats and loadRaceStats to be called first
+      await initMmrRpTimeline();
     }
 
     function gatewayChanged() {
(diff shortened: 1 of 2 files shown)


src/views/Player.vue
```diff
@@ -224,15 +224,18 @@ export default defineComponent({
...
+    async function selectSeason(season: Season): Promise<void> {
       playerStore.SET_SELECTED_SEASON(season);
-      playerStore.loadGameModeStats({});
-      playerStore.loadRaceStats();
-      playerStore.loadMatches(1);
-      playerStore.loadPlayerStatsRaceVersusRaceOnMap(battleTag.value);
-      playerStore.loadPlayerStatsHeroVersusRaceOnMap(battleTag.value);
-      playerStore.loadPlayerMmrRpTimeline();
-      playerStore.loadPlayerGameLengths();
+      await Promise.all([
```
marcoabreu, line 229: Love the parallelization!
Did you check that there's no interdependence in between those calls?

src/views/Player.vue
```diff
@@ -267,28 +270,42 @@ export default defineComponent({
       if (profile.value && battleTag.value === profile.value.battleTag) return;
 
       await playerStore.loadProfile({ battleTag: battleTag.value, freshLogin: props.freshLogin });
-      await playerStore.loadGameModeStats({});
-      await playerStore.loadRaceStats();
-      await playerStore.loadPlayerStatsRaceVersusRaceOnMap(battleTag.value);
-      await playerStore.loadPlayerStatsHeroVersusRaceOnMap(battleTag.value);
-      await playerStore.loadPlayerGameLengths();
-      await playerStore.loadMatches(1);
+      await Promise.all([
```
marcoabreu, line 273: Seems like duplicate logic. Does it make sense to centralize?

marcoabreu (approved):

src/views/Player.vue
francislavoie: As far as I could tell there wasn't any. Bunch of Ctrl+R reloading and visually comparing prod vs local and it looked correct. The only one was the MMR graph depending on `loadGameModeStats` and `loadRaceStats`

src/views/Player.vue
francislavoie: Yeah, I guess. Simple enough.

francislavoie: Done

marcoabreu (approved):

src/views/Player.vue
```diff
@@ -267,28 +263,46 @@ export default defineComponent({
...
+      ]);
+    }
+
     async function initMmrRpTimeline() {
-      const raceStats = playerStore.raceStats;
+      const selectedSeason = playerStore.selectedSeason?.id;
+      let maxMode = EGameMode.GM_1ON1;
+      let maxModeGames = 0;
+      playerStore.gameModeStats.forEach((m) => {
+        if (m.season !== selectedSeason) return;
```
gustav87, line 287: Does this guard protect against some specific scenario where the gameModeStats and selected season aren't the same?

francislavoie: Good question, I'll retest without this, I remember this fixing the result when I wrote it earlier but it may no longer be necessary after the changes to the await loading

francislavoie: You're right, wasn't needed anymore. But I did realize when clicking on old seasons, some players played some inactive modes the most, so I added a check to skip inactive modes when picking a favourite so it tends to pick one we can show an MMR graph for

src/views/Player.vue
```diff
@@ -267,28 +263,50 @@ export default defineComponent({
...
+        playerStore.loadPlayerStatsRaceVersusRaceOnMap(battleTag.value),
+        playerStore.loadPlayerStatsHeroVersusRaceOnMap(battleTag.value),
+        playerStore.loadPlayerGameLengths(),
+      ]);
+    }
+
     async function initMmrRpTimeline() {
-      const raceStats = playerStore.raceStats;
+      // Make a lookup table for active game modes
+      const activeGameModes = rankingsStore.activeModes.reduce((acc, mode) => {
```
gustav87, line 284: Is it possible to use the helper function `activeGameModes` here?
https://github.com/w3champions/website/blob/master/src/mixins/GameModesMixin.ts#L24

francislavoie: Fixed

francislavoie: Yeah okay that doesn't work actually because calling `activeGameModes()` here throws `SyntaxError: Must be called at the top of a setup function`. We already have the data loaded in the rankingsStore so we can just fetch it directly, no?

francislavoie: Yeah okay there's an issue, you're right, needs to call it since it happens in `Player.vue` but the data is normally only loaded in `PlayerStatisticTab.vue` (I've been refreshing that tab specifically so I didn't notice that it didn't load it in time

francislavoie: It's complaining on this line https://github.com/w3champions/website/blob/52dd50e06a1f29da64597cc4789e2c548b297a8d/src/mixins/GameModesMixin.ts#L71 because `useI18n()` apparently needs to be called top-level. Seems like a bug with `getGameModes()` :grimacing:

gustav87: I'll take a look in a few hours.

francislavoie: Well the bug is fixed by calling `rankingsStore.retrieveRankings()` in `Player.vue`, but I don't think we can use the mixin because of how it wants to use i18n for whatever reason. I don't understand that error. We only need the IDs here so this is enough.

gustav87: That's annoying. That's a bug I introduced when I moved all the components to the Composition API to prepare for the transition from Vue 2 to Vue 3. But I haven't continued the migration yet, because there's so much breakage related to Vuetify.

gustav87 (approved):

PR #853, opened 2024-10-26
Use classic hero icons instead of reforged

francislavoie:
Copied the classic heroes from `public/assets/raceAvatars/classic/` and renamed them to match the hero names instead of the `RACE_#.jpg` naming


src/components/overall-statistics/HeroPictureSelect.vue
```diff
@@ -156,7 +156,7 @@ export default defineComponent({
     }
 
     function parsePicture(hero: HeroPick): string {
-      return getAsset(`heroes/${hero.heroId}.png`);
```
marcoabreu, line 159: Shall we maybe centralized this function so we don't have this logic replicated in multiple places?

francislavoie: I'm not against it, but maybe in a separate PR. I'm still new here :sweat_smile:

gustav87:
I believe the images in `public\assets\raceAvatars\classic` are actually pngs, with the wrong extension. Shall we set the correct .png extension and use that instead?

francislavoie:
Oh you're right, hilarious :joy: that's easier

gustav87:
image

francislavoie:
Done, now it's just updated files, no add/delete, no code changes

gustav87 (approved):

PR #854, opened 2024-10-27
Stop using `ComputedRef` type everywhere

francislavoie:
Phew, this is a big one. Sorry. It's pretty straight-forwards though.

In TypeScript, it's generally better to lean on its inference power rather than overly-aggressively typing everything. Reduces noise, can make sure type bugs don't get accidentally hidden, etc. In places where the type isn't totally clear, using `computed<T>` instead of `ComputedRef<T> = computed((): T` is shorter and works with the type system more intentionally.

While going through it, I ran into a `v-switch` in `AdminPermissions.vue` which used both `v-model` and `:value`, which causes a type error because both of them try to set the `value` attribute. To fix that, I added `vueCompilerOptions` in `tsconfig.json` to smooth it over.

Also in `Ranking.vue` there's a really annoying type issue I can't seem to fix, because Vue 2 template don't support TypeScript. So I just left in a comment that it can be fixed in Vue 3 later on. Non-critical, but annoying in VSCode cause it's red :joy:


francislavoie:
Did a second commit, I did a pass on all `yarn lint` warnings. One of which was "components shouldn't be one word", leading to renaming the `Tournament` and `Tournaments` components.

PR #857, opened 2024-10-27
Parallelize OverallStatistics data loading, small text change

francislavoie:
Similar to #852, we can parallelize with `Promise.all`. I'm using a `.push()` approach because one of them is in an `if` so that's the most convenient way to have a variable amount of items in that array.

### DIFF
--- a/src/locales/data.ts
+++ b/src/locales/data.ts
@@ -520,7 +520,7 @@ const data = {
       America: "America",
     },
     mmrLeagueRanges: {
-      MMR_0: "all",
+      MMR_0: "All",
       MMR_2200: "> 2200 (~Grandmaster)",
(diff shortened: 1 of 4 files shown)


gustav87 (approved):

src/locales/data.ts
```diff
@@ -520,7 +520,7 @@ const data = {
       America: "America",
     },
     mmrLeagueRanges: {
-      MMR_0: "all",
+      MMR_0: "All",
```
gustav87, line 523: These things are handled in the google sheet, and will get overwritten the next time the script is run.

francislavoie: Yep, I updated it in the sheet already, just did it locally to test/confirm visually.

PR #858, opened 2024-10-27
MatchDetails: Layout adjustments, tooltips

francislavoie:
Some adjustments to the Match Details & Match History screens:

- Override the padding on the kill/exp/items boxes to shrink them a bit, so that stats for each player has more of a gap
- Vertically center the `VS` in both places (and player names on Match Details in 1v1 since the region/season box on the left causes it to be larger)
- Add the time on the match details screen (it was already shown in the match history so might as well also show it on the more detailed screen). Also added a `|` divider, thought it looked a bit more visually balanced with it. Up for debate tho.
- Increase the size of the map name/duration/time title, there's room for it and it's important information
- Changed the ping tooltip transition to slide down instead of pop in from the center, thought it felt a bit nicer
- Added the :point_up_2: cursor on the clickable area on the match history table, wasn't obvious it was clickable or where to click to get to the match details screen (`.clickable` class)
- Added a hover tooltip on the race icon, especially helpful for Random since it can be difficult to tell which race they actually got since it's behind the `?` & its drop shadow. Shows as `Random -> {RACE}` if it was random, as simply `{RACE}` otherwise for non-random. 
image
- Because of the new `v-tooltip` wrapper component, I had to rework how the margins were set, instead applying them on the player name+mmr container.
- The season translation string was `Season:` which produced `Season:: 20`. I got rid of the extra `:` in the translations (google sheet updated as well)
- Very slightly adjusted the position of the download replay button so it properly sits in the corner with even gap on both sides.

Before:
image

After:
image

Before:
image

After:
image

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -107,7 +107,7 @@
             <v-icon>{{ mdiInvertColors }}</v-icon>
           </v-btn>
         </template>
-        <v-list class="theme-selector">
+        <v-list max-height="400" class="theme-selector overflow-y-auto">
           <v-list-item @click="setTheme('human')">
             <v-list-item-title>{{ $t("races.HUMAN") }}</v-list-item-title>
           </v-list-item>
--- a/src/components/clans/MemberManagementMenu.vue
+++ b/src/components/clans/MemberManagementMenu.vue
@@ -6,7 +6,7 @@
       </v-btn>
     </template>
     <v-card>
-      <v-list dense>
+      <v-list dense max-height="400" class="overflow-y-auto">
         <v-list-item
           v-for="a in actions"
           :key="a.name"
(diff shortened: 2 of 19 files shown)


gustav87:
Great changes overall, love it. I believe we can make without the tooltip for the race icons, however. The race icons should be clear enough on their own for Warcraft 3 players.

francislavoie:
What's the harm in having the tooltips tho? I think it's especially helpful for Random because the `?` obscures the actual race quite a lot

gustav87:
Tooltips should be reserved for when needing to add extra information, but the information is already contained in the image. For random, it might make sense if the image is not clear enough, but if that's the case, I'm of the opinion it's a problem with the image, and should not be corrected with a tooltip.

francislavoie:
It's also an accessibility thing IMO, not everyone associates those icons with the race (all kinds of visual association disorders), having it spelled out when you hover is very helpful.

gustav87:
Ask kenshin about it. If Kenshin likes it, it's alright to me.

gustav87 (approved):

PR #865, opened 2024-10-29
Use map name instead of map file name in matchup statistics

CoffMakesThings:
Hi, this is my first PR here. :)

The goal is to replace these map file names with the maps' actual names.

image

The proper map names are already available from the race-on-map-versus-race endpoint.

image

race-on-map-versus-race populates the nice mapName by leveraging matchmaking-service/ladder/active-modes. If the matchmaking service does not have information about the maps in question, then mapName can be null. In that case the front end should fall back to item.map (the map file name which is currently used).

### DIFF
--- a/src/components/overall-statistics/RaceToMapStat.vue
+++ b/src/components/overall-statistics/RaceToMapStat.vue
@@ -10,8 +10,8 @@
     >
       <template v-slot:body>
         <tbody>
-          <tr v-for="item in stats" :key="item.map">
-            <td>{{ item.map }}</td>
+        <tr v-for="item in stats" :key="item.mapName || item.map">
+          <td>{{ item.mapName || item.map }}</td>
             <player-stats-race-versus-race-on-map-table-cell :stats="item.winLosses[1]" />
             <player-stats-race-versus-race-on-map-table-cell :stats="item.winLosses[2]" />
             <player-stats-race-versus-race-on-map-table-cell :stats="item.winLosses[4]" />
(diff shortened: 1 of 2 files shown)


src/components/overall-statistics/RaceToMapStat.vue
```diff
@@ -10,8 +10,8 @@
     >
       <template v-slot:body>
         <tbody>
-          <tr v-for="item in stats" :key="item.map">
-            <td>{{ item.map }}</td>
+        <tr v-for="item in stats" :key="item.mapName || item.map">
+          <td>{{ item.mapName || item.map }}</td>
```
francislavoie, line 14: The indentation change is incorrect, the `<td>` is a sibling of `<player-stats-race-versus-race-on-map-table-cell>`
```suggestion
          <tr v-for="item in stats" :key="item.mapName || item.map">
            <td>{{ item.mapName || item.map }}</td>
```

francislavoie (approved):
Looks great to me :grinning:

francislavoie:
Tested this locally, does look much better! I notice that Gloomstalker is missing `mapName` for example: `{"map":"2404091839GloomStalkerv1_1","mapName":null,"winLosses":[]}`, maybe that needs a backend data fix

francislavoie:
Before:
image

After:
image

francislavoie:
Figured I'd just merge though I'm still new here, I see nothing wrong with this! :sweat_smile:

PR #866, opened 2024-10-29
Redirect un-encoded player URLs if possible

francislavoie:
This is a bit of a fun hack.

For example, you might request https://w3champions.com/player/Ruhigengeist#1200 but this results in a broken page since no user is found. We can be smarter about this because the `#` is the "fragment" (aka "hash") part of the URL, so we can redirect on mount to the correct URL i.e. https://w3champions.com/player/Ruhigengeist%231200 and it can even work if there's a path like `/matches` at the end.

### DIFF
--- a/src/views/Player.vue
+++ b/src/views/Player.vue
@@ -244,9 +244,9 @@ export default defineComponent({
     onMounted((): void => {
       // Since # is a reserver character for the "fragment/hash" part of the URL
       // we can try to redirect to the correct URL if the hash is a battle tag,
-      // (starts with 4 digits). It may also be followed by a tab name, for
-      // example Someone#9999/matches would redirect to Someone%239999/matches
-      if (window.location.hash.match(/^#\d{4}\b/)) {
+      // (starts with at least 4 digits). It may also be followed by a tab name,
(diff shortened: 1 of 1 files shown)


src/views/Player.vue
```diff
@@ -240,6 +240,19 @@ export default defineComponent({
...
     }
 
+    onMounted((): void => {
+      // Since # is a reserved character for the "fragment/hash" part of the URL
+      // we can try to redirect to the correct URL if the hash is a battle tag,
+      // (starts with at least 4 digits). It may also be followed by a tab name,
+      // for example Someone#9999/matches would redirect to Someone%239999/matches
+      if (window.location.hash.match(/^#\d{4}/)) {
+        const url = new URL(window.location.href);
+        url.pathname = `${url.pathname}${url.hash}`;
```
marcoabreu, line 250: And this will automatically be escaped when we copy over the # ?

francislavoie: Yep! Browser magic, automatically re-encodes.

gustav87 (approved):

gustav87:
Thanks! Could you rebase and push again?

marcoabreu (approved):

PR #870, opened 2024-11-26
Improve filter and order on the matches page for ongoing and finished

sivertsenstian:
## Current

image

image
map doesn't seem to work here and filled with a lot of noise
image

## Updated
image

image

Added sort order in addition to column
- starttime and mmr for ongoing
- endtime and mmr for finished

Also cleaned up the map list to only show maps that have a valid map name (handled by mmr service I think, this was also improved in a recent PR that hasn't been merged yet - should work for newer seasons at least)
image

Now uses the mapName property to display in ui as well as filtering in backend

This makes it possible to find finished games on a given map
image

Or just recent high lvl games with the mmr descending sort
image

### DIFF
--- a/components.d.ts
+++ b/components.d.ts
@@ -122,6 +122,7 @@ declare module 'vue' {
     SeasonSelect: typeof import('./src/components/common/SeasonSelect.vue')['default']
     SignInDialog: typeof import('./src/components/common/SignInDialog.vue')['default']
     SocialBox: typeof import('./src/components/common/SocialBox.vue')['default']
+    SortDirectionSelect: typeof import('./src/components/matches/SortDirectionSelect.vue')['default']
     SortSelect: typeof import('./src/components/matches/SortSelect.vue')['default']
     StreamedMatchInfo: typeof import('./src/components/matches/StreamedMatchInfo.vue')['default']
     StreamedMatchPlayerInfo: typeof import('./src/components/matches/StreamedMatchPlayerInfo.vue')['default']
--- a/src/components/common/MapSelect.vue
+++ b/src/components/common/MapSelect.vue
@@ -15,9 +15,9 @@
         </v-list>
         <v-divider></v-divider>
         <v-list dense max-height="400" class="overflow-y-auto">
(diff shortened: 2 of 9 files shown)


francislavoie:
Looks good! Did you make the changes to the translation google sheet?

sivertsenstian:
> Looks good! Did you make the changes to the translation google sheet?

No - there was a data.ts file with some translations, I just added/expanded on the English one there

francislavoie:
That file is generated from https://docs.google.com/spreadsheets/d/1V5f4zguWDmk9nbnoXSJm9g-ZxImo83NJpSY17EUkzOc/edit#gid=0, you can request access to edit. In prod it doesn't read from `data.ts` so it needs to be ready ahead of time :+1:

gustav87:
It was a long time ago now, so I don't remember clearly, but this endpoint disabled at first, then it was enabled without any filters, because we experienced db performance issues. The query was very expensive. I recommend removing the mmr filtering feature. Are there indexes to support the sortings you have enabled?

gustav87:
I remember now that we added the season parameter, so the db no longer has to process all 10 million records, so it should probably be better now. But the mmr range filter could possibly still be expensive for finished matches.

sivertsenstian:
> It was a long time ago now, so I don't remember clearly, but this endpoint disabled at first, then it was enabled without any filters, because we experienced db performance issues. The query was very expensive. I recommend removing the mmr filtering feature. Are there indexes to support the sortings you have enabled?

I think you mentioned something about there not being a season filter included previously? 

I have no idea about indexes, it uses the same sorting from before, but now also for finished games - I see the document type is different so not sure, how do I check this? 

Is it possible to get a production data dump from season 19/20 to test performance?

gustav87:
Finished and ongoing matches are stored in different collections, so a sorting of the ongoing matches does not necessarily perform well on finished matches, depending on the indexes present.

I will send you a prod dump on discord.

gustav87:
You can add any kind of filter and sorting on ongoing matches, because the collection is tiny. For finished matches, it takes a bit more thought, because of the millions of documents present.

PR #878, opened 2025-04-04
display error message if player profile not found

gustav87:


francislavoie:
https://github.com/w3champions/website/pull/866 is somewhat related. If you load the page with `#` instead of `%26` it would fix that with a redirect.

marcoabreu (approved):

gustav87:
> #866 is somewhat related. If you load the page with `#` instead of `%23` it would fix that with a redirect.

I had missed that one. Thanks. Approved!

PR #879, opened 2025-04-10
Dedupe FFA list, fix v-model on computed property warning

francislavoie:
We had the list of FFA modes in multiple places, figuring consolidating makes sense.

Noticed Vue was warning for the selects on match history due to the computed properties being used as a model, should just bind the value instead.


src/store/constants.ts
```diff
@@ -0,0 +1,8 @@
+import { EGameMode } from "./types";
+
+export const GAME_MODES_FFA = [
+    EGameMode.GM_FFA,
+    EGameMode.GM_LTW_FFA,
```
gustav87, line 5:
```suggestion
    EGameMode.GM_FFA,
```

gustav87: LTW is no longer an active ladder. We keep the game mode in the matchmaking code base if we ever want to enable it again. Then we'll add it back to website and website-backend.

francislavoie:
Thanks!

PR #880, opened 2025-04-14
Adds Hero Icons to matches view

Magentah:
I'm not a front end dev or capable at all with UI, so there are quite likely issues or obvious improvements that could be made for this. I added the icons, hopefully, in a way that doesn't take up too much space and works okay for the matches grid. I defaulted the icons to be hidden by default, and also to not have the icons show the level as they take up a lot of space.

- `HeroIcon` component updated to have a toggle for displaying level or not as showing the level flag in the matches grid didn't look good as it took up a lot of space. 
- Added `HeroIconRow` component for displaying the hero icons in the Matches view
- Added `HeroIconToggle` component to toggle displaying hero icons and not.
    - Also updated `MatchesGrid` to pass toggle value
    - Defaults to hidden
- Some changes to padding/margins

Image to show what it looks like
image

### DIFF
--- a/src/components/matches/MatchesGrid.vue
+++ b/src/components/matches/MatchesGrid.vue
@@ -4,11 +4,7 @@
       <table class="custom-table">
         <thead>
           <tr>
-            <td
-              v-for="header in headers"
-              :key="header.name"
-              :style="header.style"
-            >
+            <td v-for="header in headers" :key="header.name" :style="header.style">
               {{ header.text }}
             </td>
             <td v-if="!unfinished">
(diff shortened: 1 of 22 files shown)


Magentah:
The tooltips are already there. It's not *too* bad for more than 1v1, but it does take up a lot of space.

I wasn't sure about the size for the icons. Too small and I think they're not readable so they're not particularly useful. I'm also awful at styling so I don't actually know how move the icons to directly under the names, adjust the race icon, and have it all work and look reasonable. I agree it'd probably look better though.

francislavoie:
I'll gladly touch up the UI when I can 👍 I haven't figured out how to run the backend though so I'd have to do some annoying mocking 😅

francislavoie:
Yeah, much better! Thanks for adjusting 😊

src/components/matches/MatchesGrid.vue
```diff
@@ -164,6 +168,11 @@ export default defineComponent({
       type: Boolean,
       required: true,
     },
+    showHeroes: {
+      type: Boolean,
+      required: false,
+      default: true,
```
marcoabreu, line 174: How about we call it "show details" and it's off by default in order to keep the history clean, but allow users to expand if they want

src/views/Matches.vue
```diff
@@ -76,6 +81,8 @@ export default defineComponent({
     const map = computed<string>(() => matchStore.map);
     const mmr = computed<Mmr>(() => matchStore.mmr);
 
+    const showHeroIcons = ref<boolean>(true);
```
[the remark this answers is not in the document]
Magentah: [This v-for](https://github.com/w3champions/website/pull/880/files/7292e01a6d70216a0a6edfbd0c5fa61bafb52042#diff-e4560d015ca612c72466a24041b80a09122a25c1bc5811304808f1d196fe14c7R3) handles no data. If the data doesn't exist, it just doesn't show anything.

src/components/matches/MatchesGrid.vue
Magentah: Isn't `showDetails` less specific than `showHeroes`, when it's only being used specifically for showing hero icons? There are no other details. If there are more details added later, we can change it then?

marcoabreu: Fine with me :)

src/store/types.ts
```diff
@@ -80,7 +81,7 @@ export interface UnitScore {
 }
 
 export interface Hero {
-  icon: string;
+  name: string;
```
marcoabreu, line 84: You're identifying heros by their name in the frontend anyways?

Magentah: For the one I know yeah, I wasn't sure if other things use the existing `icon` field and didn't want to break them.

marcoabreu: Shouldn't we rather be using IDs?

BogdanW3: I would also say this struct will need to have some way to discriminate maps - if we go with ids, there could be multiple heroes with id H001, if we go with icon paths, multiple different icons could be Texture/Hero.dds

Magentah: The name is used for translation keys and paths to images, which make sense to be names to me. I changed to `name` instead of `icon` as I thought it's clearer, as it's not just for the icon.

Either way, there's no reason why `name` can't include a prefix of some kind if we want later which can be used across both asset path and translation keys.

Magentah: The backend has been updated to also have a `HeroSource` field for discriminating heroes

marcoabreu (approved):

PR #881, opened 2025-04-14
Preserve sorting when rankings update

CoffMakesThings:
From Neo

idk where to put this, but i might have a feature / bugfix request 😅 

when i'm on the homepage https://www.w3champions.com/rankings
and sort by MMR, the sorting resets after x seconds
is that intended or something we can fix?


CoffMakesThings:
I think the updated data needs to be sorted again. Will rethink

PR #888, opened 2025-05-01
Adds gateway America

### LINKED ISSUE #887: Sign-in dialog lists EU and China gateways, but not America
The sign-in dialog (https://w3champions.com/player/?freshLogin=true) lists Europe and China gateways, but not America. To login, I had to select Europe and then later change the gateway to America.

Image

amarvin:
Need to:

- [x] confirm that battle.net can be used for authentication in America
- [x] add `public/assets/flags/us.svg`

Closes #887. Builds off #388.


francislavoie:
It doesn't actually make a difference if americans use the european gateway or not, the end result is the same. In the launcher, we recently changed "Europe" to "Global" because it has a better meaning for what is actually happening, which is just authenticating with battlenet's common auth system.

amarvin:
Ah, good point. Maybe I can just change this PR to simply rename it to Global then.

src/components/common/SignInDialog.vue
```diff
@@ -62,6 +62,11 @@ export default defineComponent({
     const oauthStore = useOauthStore();
 
     const gateways = [
+      {
+        id: BnetOAuthRegion.us,
+        name: t("gatewayNames.America"),
+        uri: "https://battle.net",
```
amarvin, line 68: Maybe the correct link is:

```suggestion
        uri: "https://us.battle.net",
```

I learned that from this comment: https://github.com/w3champions/launcher/pull/77#issuecomment-765263822.

src/store/oauth/types.ts
```diff
@@ -24,6 +24,7 @@ export interface TwitchToken {
 }
 
 export enum BnetOAuthRegion {
+  us = "us",
```
marcoabreu, line 27: We will have to check if this has any downstream implications if the session is established via another gateway now.

francislavoie: I think it's certainly less risky to leave this as-is and call it global, we've already effectively established that the Europe gateway is what we use globally.

francislavoie:
Superseded by https://github.com/w3champions/website/pull/956

PR #890, opened 2025-05-23
Add global mute search query, nextId and created at properties

gustav87:


src/services/admin/AdminService.ts
```diff
@@ -62,8 +62,14 @@ export default class AdminService {
...
   }
 
-  public static async getGlobalMutes(token: string): Promise<GloballyMutedPlayer[]> {
-    const url = `${API_URL}api/admin/globalChatBans`;
+  public static async getGlobalMutes(token: string, searchQuery: string | undefined, nextId: number | undefined): Promise<GloballyMutedPlayer[]> {
+    let url = `${API_URL}api/admin/globalChatBans`;
+    if (searchQuery) {
+      url += `?query=${searchQuery}`;
+    }
+    if (nextId) {
```
marcoabreu, line 70: Can searchQuery and nextId be set at the same time?

src/store/admin/types.ts
```diff
@@ -84,6 +84,7 @@ export interface GlobalMute {
 
 export interface GloballyMutedPlayer extends GlobalMute {
   id: string;
+  createdAt: string;
```
marcoabreu, line 87: How come this is part of GloballyMutedPlayer and not GlobalMute?

gustav87: Because it's not known at the time of creation

src/components/admin/AdminGlobalMute.vue
```diff
@@ -158,6 +158,7 @@ export default defineComponent({
 
     const globallyMutedPlayers = computed<GloballyMutedPlayer[]>(() => adminStore.globallyMutedPlayers);
     const banDateSet = computed<boolean>(() => banExpiry.value != "");
+    const author = computed<string>(() => oauthStore.blizzardVerifiedBtag);
```
marcoabreu, line 161: What is this here for?

Lets make sure we retrieve and set the author on the website-backend; rather than trusting the input - just to avoid impersonation.

gustav87: Permission checking is done in the backend already. We trust one mod to not impersonate another.

marcoabreu: But why even bother introducing this as a property in the frontend if we can just retrieve it in the backend?

If we have the ability to retrieve it from the auth credentials, I'd definitely prefer that. Verify > trust.

Issue #899, opened 2025-06-24
Show alias on rankings page


marcoabreu:
We already have alias support for profiles:

![Image](https://github.com/user-attachments/assets/c9c574d9-5894-4287-8653-b0f7a31bf86d)

But we do not have it on the rankings. Maybe we could show the alias there:

![Image](https://github.com/user-attachments/assets/1970ffc4-a1e9-4e57-9b7e-afc969222f6e)

PR #903, opened 2025-07-17
add alias to ranking page

quan194:
add alias to ranking page https://github.com/w3champions/website/issues/899

### DIFF
--- a/src/components/ladder/RankingsGrid.vue
+++ b/src/components/ladder/RankingsGrid.vue
@@ -46,7 +46,8 @@
                   'background-image': 'url(' + getRaceIcon(item, index) + ')',
                 }"
               ></div>
-              <player-rank-info :player-id="playerId" />
+              <player-rank-info :player-id="playerId"
+                :alias="item.playersInfo[index].playerAkaData.name" />
               <div
                 class="ml-1"
                 v-if="(item.playersInfo && item.playersInfo[index].countryCode) || item.playersInfo[index].location"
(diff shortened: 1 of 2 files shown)


src/components/ladder/RankingsGrid.vue
```diff
@@ -46,7 +46,8 @@
                   'background-image': 'url(' + getRaceIcon(item, index) + ')',
                 }"
               ></div>
-              <player-rank-info :player-id="playerId" />
+              <player-rank-info :player-id="playerId"
+                :alias="item.playersInfo[index].playerAkaData.name ? item.playersInfo[index].playerAkaData.name : null" />
```
CoffMakesThings, line 50: In website-backend we have a type PlayerInfo
`public Player PlayerAkaData { get; set; }`
This means that it will never be null. 
`public string name { get; set; } // The anglicised nickname of the player by the community (their aka)`
This one is also never null. If they could be null they would have a question mark, like this:
public int? Color { get; set; }

Therefore I don't think you need any null check. Could you also please modify the PlayerInfo interface on the front end so that it includes playerAkaData?

src/components/ladder/RankingsGrid.vue
```diff
@@ -509,12 +510,15 @@ export default defineComponent({
 
 <style lang="scss" scoped>
 .theme--light {
+
```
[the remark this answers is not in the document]
CoffMakesThings: I mean these line breaks in the css

src/components/ladder/RankingsGrid.vue
CoffMakesThings: But based on the code that generates these PlayerInfos, I'm pretty confident it will never be null

CoffMakesThings: In summary since I went on a confusing rant, I think it should be:
:alias="item.playersInfo[index].playerAkaData.name" />
or as suggested by MicroHive
:alias="item.playersInfo[index].playerAkaData?.name" />
Up to you

quan194: i fixed it to become :alias="item.playersInfo[index].playerAkaData.name" />

PR #929, opened 2025-09-09
Swap out yarn for npm

gustav87:
Npm was replaced with yarn in May of 2022, before I joined the project. Not sure about the reasons for it, but I imagine it sped up the build process.

Nowadays npm is faster and yarn is not as popular, and I'm a bit confused by the many different (incompatible?) versions of yarn, so I'm switching it back to npm.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 # build stage
 FROM node:20-bullseye as build-stage
 WORKDIR /app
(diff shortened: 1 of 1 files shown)


gustav87:
> I agree, npm is a better default now than yarn.

I was just about to add you as a reviewer :) Thanks!

PR #955, opened 2025-09-27
Refactor "Setup Guides" into FAQ page

francislavoie:
- Moved the Setup Guides to be a subtab of the FAQ section. This removes an item from the header bar, saving some space there, and consolidating similar "information" pages into one.
- Added router redirects for old paths to the new paths below `/faq` (Download link on the homepage still works via the redirect)
- Adjusted styles a bit (mostly padding) for the Launcher Setup screen
- Improved the "Installing WarCraft 3" instructions with more detailed steps and updated screenshots, covering the whole "logout then continue without logging in" flow which is awkward a f but it's what needs to be done for now since Blizzard hasn't fixed that launcher quirk yet.
- I've already added the translations to the google spreadsheet

image

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -9,7 +9,7 @@
       <v-list dense>
         <v-list-item>
           <v-list-item-content>
-            <router-link :to="{ name: 'Home' }">
+            <router-link :to="{ name: EMainRouteName.HOME }">
               <brand-logo :is-dark-theme="isDarkTheme.get()" />
             </router-link>
           </v-list-item-content>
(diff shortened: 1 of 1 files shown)


francislavoie:
FYI @jeremyduez since you worked on this last

francislavoie:
Ah, my apologies @jeremyduez I wasn't sure where you were at on it and it was still bothering me a bit so I went ahead and did it, had the time today.

gustav87 (approved):
Very nice! :)

jeremyduez:
Absolutely no problem @francislavoie, thx for doing it well!

PR #960, opened 2025-09-28
Fix MMR select visibility logic in Matches view

godrien:
This pull request makes a small change to the `Matches.vue` view by always rendering the `mmr-select` component, regardless of whether the match is unfinished. Previously, `mmr-select` was only shown for unfinished matches. This simplifies the component's visibility logic.

- Always render the `mmr-select` component by removing the `v-if="unfinished"` condition in `Matches.vue`.


godrien:
Yes, backend support for that was added here https://github.com/w3champions/website-backend/pull/474

PR #961, opened 2025-10-07
Vuetify3 theme

gustav87:
Making the theme styling work. I'm not sure I'm doing it right, and it's still a work in progress, but this brings it a bit more in line with prod anyway.


gustav87:
Scratch that. I added the `success` color to the themes, and then vuetify automatically adds lighten and darken for it so I could use that for orc and human.

It still feels like I'm doing it wrong, but whatever. What have you done in terms of the themes? Can we merge them somehow? Shall I scratch mine?

francislavoie:
No, your changes are good, keep going 👍

PR #1024, opened 2026-04-10
Display timestamps on admin-view-game-chat

CoffMakesThings:
image

### DIFF
--- a/src/components/admin/replays/ReplayChatMessage.vue
+++ b/src/components/admin/replays/ReplayChatMessage.vue
@@ -66,7 +66,7 @@ export default defineComponent({
     // Example: 1500     -> 00:15
     // Example: 15000000 -> 04:10:00
     function formatTime(timeMs: number): string {
-      const totalSeconds = Math.max(0, Math.floor(timeMs / 1000));
+      const totalSeconds = Math.floor(timeMs / 1000);
       const hours = Math.floor(totalSeconds / 3600);
       const minutes = Math.floor((totalSeconds % 3600) / 60);
       const seconds = totalSeconds % 60;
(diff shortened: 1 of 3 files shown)


src/components/admin/replays/ReplayChatMessage.vue
```diff
@@ -60,12 +63,36 @@ export default defineComponent({
       }
     }
 
+    // Example: 1500     -> 00:15
+    // Example: 15000000 -> 04:10:00
+    function formatTime(timeMs: number): string {
+      const totalSeconds = Math.max(0, Math.floor(timeMs / 1000));
```
gustav87: Do you need the `Math.max()` here?

CoffMakesThings: Nah, I think we can assume timeMs will always be 0 or more. I'll remove the Math.max

src/components/admin/replays/ReplayChatMessage.vue
```diff
@@ -60,12 +63,36 @@ export default defineComponent({
...
+      const hours = Math.floor(totalSeconds / 3600);
+      const minutes = Math.floor((totalSeconds % 3600) / 60);
```
gustav87, line 71: Just for info, you could use the `date-fns` lib we're already using on the website for date/duration related things. https://date-fns.org/v4.1.0/docs/millisecondsToMinutes

gustav87 (approved):

src/components/admin/replays/ReplayChatMessage.vue
CoffMakesThings: I'll try to remember that lib for next time

gustav87: For this case maybe not so necessary, but it has a lot of good utility functions.

PR #1025, opened 2026-04-11
simplify topGameModeStats function. filter out duplicate stats

gustav87:
Before:
image

After:
image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -50,6 +50,7 @@
       },
       "devDependencies": {
         "@eslint/js": "^9.39.1",
+        "@types/lodash": "^4.17.24",
         "@types/node": "^20.11.30",
         "@vitejs/plugin-legacy": "^8.0.1",
         "@vitejs/plugin-vue": "^6.0.5",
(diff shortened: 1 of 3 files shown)


CoffMakesThings (approved):
The old code would prioritize melee 1v1 and melee 2v2, as your first two displayed modes. The new code prioritizes whatever you are best ranked at. Some players might be surprised to see that they now have Legion TD modes displayed on their profile, instead of 1v1 and 2v2, because they are better ranked at Legion TD.

The ideal would be that people can pick whatever they want but that is more complicated.

gustav87:
> The old code would prioritize melee 1v1 and melee 2v2, as your first two displayed modes. The new code prioritizes whatever you are best ranked at. Some players might be surprised to see that they now have Legion TD modes displayed on their profile, instead of 1v1 and 2v2, because they are better ranked at Legion TD.
> 
> The ideal would be that people can pick whatever they want but that is more complicated.

I think that was the logic in the past, but that doesn't seem to hold true anymore. For example:
Aware: https://w3champions.com/player/Aware%2311107 or
Shocker Season 20: https://w3champions.com/player/Shocker%2321685

We could do that if we want to though. We could also display multiple of the same mode if there is there room for them in the UI without pushing out other modes. Jens could have both his races for 1v1 displayed, for instance: https://w3champions.com/player/mayhem%2312806

CoffMakesThings:
> I think that was the logic in the past, but that doesn't seem to hold true anymore

Right, my bad, I see now that on line 165 of the old code we are anyway putting all the modes together and sorting by rank. So this change is doing what you described - Simplifying the implementation of the existing behaviour and removing those duplicates.

gustav87:
No worries, I was confused as well.

Issue #1023, opened 2026-04-10
Add Duration Range Filter to Finished Games Search


MichaelDeitz:
The backend now supports filtering finished matches by minimum and maximum duration (in seconds).
The website UI should expose this filter so users can search for games within a specific duration range.

Backend Work (already completed)
New Query Parameters
The /api/matches endpoint now accepts:
- minDuration — minimum match duration in seconds
- maxDuration — maximum match duration in seconds

Model Field Used
The backend filters on:
Matchup.Duration (TimeSpan)
which maps to the stored durationInSeconds field in MongoDB

Here is where it is expected to be.

<img width="1286" height="436" alt="Image" src="https://github.com/user-attachments/assets/01832939-bcdd-4f7c-9546-b376b60cf7fe" />

Frontend Work Needed
Add two numeric inputs for:
- Min Duration (seconds)
- Max Duration (seconds)
These should be placed in the Matches filter panel, in the same section as the MMR range filters (see screenshot).
When values are entered, the frontend should append them to the API request:
minDuration=<value>
maxDuration=<value>

PR #1029, opened 2026-04-16
Duration filter fix

MichaelDeitz:
Duration Filter (#1023)

A new Duration filter has been added to improve how users refine results based on time.

The filter appears as a “Duration” button.
When selected, it opens a scrollable range selector allowing users to choose a duration between 0 minutes and 24 hours.
Users can easily adjust the range by scrolling to their desired time window.

Resetting the filter:

To remove the applied duration filter, scroll the range back to 0–24 hours.
Once reset, the button will revert to displaying “Duration”, indicating that no filter is currently applied.

### DIFF
--- a/src/services/MatchService.ts
+++ b/src/services/MatchService.ts
@@ -12,16 +12,25 @@ export default class MatchService {
     gameMode: EGameMode,
     map: string,
     mmr: Mmr,
-    season: number,
+    duration: { min: number; max: number },
+    season: number | { id: number } | null | undefined,
     heroes: number[],
   ): Promise<{ count: number; matches: Match[] }> {
     const offset = page * this.pageSize;
     const minMmr = mmr.min === 0 ? "" : `&minMmr=${mmr.min}`;
     const maxMmr = mmr.max === 3000 ? "" : `&maxMmr=${mmr.max}`;
-    const heroQuery = Array.isArray(heroes) && heroes.length > 0
-      ? heroes.filter((h) => h > 0).map((h) => `&hero=${h}`).join("")
-      : "";
-    const url = `${API_URL}api/matches?offset=${offset}&gateway=${gateway}&pageSize=${this.pageSize}&gameMode=${gameMode}&map=${map}${minMmr}${maxMmr}&season=${season}${heroQuery}`;
+    const minDuration = duration.min > 300 ? `&minDuration=${duration.min}` : "";
+    const maxDuration = duration.max < 99999 ? `&maxDuration=${duration.max}` : "";
+    const heroQuery = Array.isArray(heroes) && heroes.length > 0 ? heroes.filter((h) => h > 0).map((h) => `&hero=${h}`).join("") : "";
+    let seasonParam: number = -1;
+
+    if (typeof season === "number") {
+    seasonParam = season;
+  } else if (typeof season === "object" && season !== null && "id" in season) {
+    seasonParam = (season as { id: number }).id;
+  }
+
+    const url = `${API_URL}api/matches?offset=${offset}&gateway=${gateway}&pageSize=${this.pageSize}&gameMode=${gameMode}&map=${map}${minMmr}${maxMmr}${minDuration}${maxDuration}&season=${seasonParam}${heroQuery}`;
     const response = await fetch(url);
     return await response.json();
   }
@@ -32,6 +41,7 @@ export default class MatchService {
     gameMode: EGameMode,
     map: string,
     mmr: Mmr,
+    duration: { min: number; max: number },
     sort: string,
   ): Promise<{ count: number; matches: Match[] }> {
     const offset = page * this.pageSize;
@@ -43,6 +53,7 @@ export default class MatchService {
       gameMode,
       map,
       mmr,
+      duration,
       sort,
     );
   }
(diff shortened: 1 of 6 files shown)


francislavoie:
You'll need to rebase this, you can drop your `package-lock.json` changes

src/services/MatchService.ts
```diff
@@ -54,11 +70,16 @@ export default class MatchService {
...
     sort: string,
   ): Promise<{ count: number; matches: Match[] }> {
     const minMmr = mmr.min === 0 ? "" : `&minMmr=${mmr.min}`;
     const maxMmr = mmr.max === 3000 ? "" : `&maxMmr=${mmr.max}`;
-    const url = `${API_URL}api/matches/ongoing?offset=${offset}&gateway=${gateway}&pageSize=${pageSize}&gameMode=${gameMode}&map=${map}${minMmr}${maxMmr}&sort=${sort}`;
+    const _durationQuery = duration
+      ? `&minDuration=${duration.min}&maxDuration=${duration.max}`
+      : "";
+
+    const url = `${API_URL}api/matches/ongoing?offset=${offset}&gateway=${gateway}&pageSize=${pageSize}&gameMode=${gameMode}&map=${map}${minMmr}${maxMmr}${_durationQuery}&sort=${sort}`;
```
CoffMakesThings: This is adding a duration parameter to ongoing matches. There is no duration param for ongoing matches on the BE.

So, I think the call to this endpoint should not be modified in this PR.

src/store/match/store.ts
```diff
@@ -18,6 +18,7 @@ export const useMatchStore = defineStore("match", {
     gameMode: EGameMode.GM_1ON1,
     map: "Overall",
     mmr: { min: 0, max: 3000 } as Mmr,
+    duration: null as { min: number; max: number } | null,
```
CoffMakesThings: Can this be instantiated with the default values, like mmr is, instead of null?

src/services/MatchService.ts
```diff
@@ -12,16 +12,30 @@ export default class MatchService {
     gameMode: EGameMode,
     map: string,
     mmr: Mmr,
-    season: number,
+    duration: { min: number; max: number } | null,
+    season: number | { id: number } | null | undefined,
```
CoffMakesThings: The type of season is Season.

src/services/MatchService.ts
```diff
@@ -12,16 +12,30 @@ export default class MatchService {
...
+      ? `&minDuration=${duration.min}&maxDuration=${duration.max}`
+      : "";
+
     const heroQuery = Array.isArray(heroes) && heroes.length > 0
       ? heroes.filter((h) => h > 0).map((h) => `&hero=${h}`).join("")
       : "";
-    const url = `${API_URL}api/matches?offset=${offset}&gateway=${gateway}&pageSize=${this.pageSize}&gameMode=${gameMode}&map=${map}${minMmr}${maxMmr}&season=${season}${heroQuery}`;
+
+    let seasonParam: number = -1;
+    if (typeof season === "number") {
```
CoffMakesThings: The type of season is Season so there's no need for this if condition

CoffMakesThings:
flicker

would be nice to fix this flickering

src/services/MatchService.ts
```diff
@@ -12,16 +12,30 @@ export default class MatchService {
     gameMode: EGameMode,
     map: string,
     mmr: Mmr,
-    season: number,
+    duration: { min: number; max: number } | null,
```
CoffMakesThings: Does duration need to be nullable? I think no.

src/store/match/store.ts
```diff
@@ -51,7 +53,8 @@ export const useMatchStore = defineStore("match", {
           this.gameMode,
           this.map,
           this.mmr,
-          this.selectedSeason.id,
+          this.duration,
+          this.selectedSeason,
```
CoffMakesThings: We used to pass selectedSeason.id, but now we pass selectedSeason. This is why the type of season in MatchService has been changed from number to Season. It is better to leave it as selecteSeason.id imo, and keep the type as number.

src/views/Matches.vue
```diff
@@ -11,6 +11,7 @@
             <game-mode-select :disabledModes="disabledGameModes" :gameMode="gameMode" @gameModeChanged="gameModeChanged" />
             <map-select :mapInfo="maps" :map="map" @mapChanged="mapChanged" />
             <mmr-select :mmr="mmr" @mmrFilterChanged="mmrFilterChanged" />
+            <duration-select :duration="duration" @durationFilterChanged="durationFilterChanged" />
```
CoffMakesThings: duration-select should only be rendered when we are looking at Finished matches

src/services/MatchService.ts
MichaelDeitz: Will removed

src/views/Matches.vue
MichaelDeitz: <duration-select v-if="!unfinished" :duration="duration" @durationFilterChanged="durationFilterChanged" /> does this look better?

src/store/match/store.ts
MichaelDeitz: will fix

src/services/MatchService.ts
MichaelDeitz: will fix

src/services/MatchService.ts
MichaelDeitz: let seasonParam: number = -1;
    if (typeof season === "number") {
      seasonParam = season;
    } else if (typeof season === "object" && season !== null && "id" in season) {
      seasonParam = (season as { id: number }).id;
    Do you want this whole part removed?

src/views/Matches.vue
CoffMakesThings: Yeah

src/services/MatchService.ts
CoffMakesThings: Yes

src/store/match/store.ts
MichaelDeitz: fixed

src/services/MatchService.ts
MichaelDeitz: fixed

src/views/Matches.vue
MichaelDeitz: fixed

src/services/MatchService.ts
MichaelDeitz: fixed

MichaelDeitz:
Should all be changed. Had to move away from the scroll bar as it was still jumpy after all the changes.

CoffMakesThings:
Tried it locally and it worked fine
image

CoffMakesThings (approved):
Good job

PR #1030, opened 2026-04-18
lag reports styling and UX improvements

francislavoie:
Tons of little things. Hard to remember everything.

- Fixed up CLAUDE.md, it was still ref-ing the wrong Vue/Vuetify versions, it still thought we use yarn, but we're on npm now.
- When into a report and back out, it used to reset the filters and reload the data. Now it keeps the filters and the data (but still attempts a refresh of the data).
- Added a refresh results button so if you're sitting on the page for a while waiting for stuff to come in, you can just click it instead of refreshing the whole page.
- Graph limits max zoom-out, it was annoying to use tbh.
- Turned off graph animations, more responsive.
- Completely redid the graph hover tooltips, looks way better.
- When you hover on a line, it fades out all the other lines so you can focus on the one you're hovered on. Looks sweet too.
- Changed bar graphs for lag to points, the bar graphs looked insane.
- Removed the stacking transparent/opaque BGs on tables/cards/etc, looks cleaner when the actual theme bg shows up. This'll apply on all pages, I _think_ it should be fine.
- Changed all the chips (little labels) to use "tonal" colors throughout, I thought it looked best and most consistent with best contrast. Orc theme still looks like shit because it's too yellow/tan and nothing looks good on that color contrast-wise. I can't fix it. Human and both dark themes look fine I think, but the warning/orange text is still a bit hard to read on light themes. Bleh. Sorry :man_shrugging: 
- Chips with a player name shows the player color as well in most cases (might've missed some).
- Point-in-time inspector buttons changed to clickable chips, more consistent with the events at the top and has better color
- The tab list is not all-uppercase, and uses player colors.

I think that's about it.


src/scss/shared/overrides.scss
```diff
@@ -26,6 +26,57 @@ h3 {
...
+.v-card {
+  .v-card {
+    background-color: transparent !important;
+  }
+}
+
+// Make tonal chip backgrounds opaque by pre-multiplying the color instead of using opacity.
+// This prevents the chip background from blending with whatever is behind it.
+// Dark themes blend toward black; light themes blend toward white.
+.v-theme--nightelf,
```
gustav87: These theme specific overrides should go in their respective theme scss file, perhaps. `src\scss\dark\nightelf.scss` for example

gustav87 (approved):

PR #1031, opened 2026-04-18
remove package `@vitejs/plugin-legacy`

gustav87:
For some reason, building for legacy browsers impacts the rendering also in modern browsers. I'm not sure why, but it seems like a bug to me.

If supporting legacy browsers comes at the cost of impacting the rendering of modern browsers, I'd rather remove it, because it makes development more difficult when you in your development flow also need to build the site to verify how things look.

This text field, for instance, is not supposed to be padded left when using the `underlined` type of field.

When running `npm start`:
image

When running `npm preview` or prod:
image

I first thought it was because of the manual chunking we're doing when building the site, but removing that had no effect. I also tried tinkering with some `@vitejs/plugin-legacy` options, but didn't find anything relevant.

I suspect we don't have a lot of visitors to w3champions who aren't able to update their browser, anyhow. For most individuals, this happens automatically.

Reference:
https://vuetifyjs.com/en/components/text-fields

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -52,7 +52,6 @@
         "@eslint/js": "^9.39.1",
         "@types/lodash": "^4.17.24",
         "@types/node": "^20.11.30",
-        "@vitejs/plugin-legacy": "^8.0.1",
         "@vitejs/plugin-vue": "^6.0.5",
         "@vue/eslint-config-typescript": "^14.6.0",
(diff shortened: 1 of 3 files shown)


francislavoie (approved):
Yep, legacy browsers are not worth supporting anymore for non-corporate type of websites

PR #1034, opened 2026-04-25
Reset player match filters on profile change

drakonli:
## Summary

Resets the player match filters when the selected profile changes.

This prevents filters from carrying over between different player profiles and showing stale or confusing match results.

## Changes

- Clear match filters when navigating to a different player profile
- Keep the match list state aligned with the currently selected player
- Avoid stale filter state affecting newly loaded profile data

## Testing

- Verified locally by switching between player profiles and confirming the match filters reset as expected


gustav87 (approved):

gustav87:
Looks good to me!

PR #1035, opened 2026-04-25
Add inline match details to player history

drakonli:
## Summary

Adds inline match details to player match history.

## Changes

- Extracts match detail rendering into `MatchDetailContent.vue` so it can be reused by both the standalone match page and inline match history rows.
- Enables expandable inline match details in `PlayerMatchesTab`.
- Updates `MatchesGrid` so finished match rows can expand inline while preserving:
  - player name navigation
  - match page button behavior
  - replay download behavior
- Adds Match/Open Match localization labels.

## Verification

- `npm run lint`
- `npm run build`
- local manual testing

Screenshot_3

### DIFF
--- a/src/components/match-details/MatchDetailContent.vue
+++ b/src/components/match-details/MatchDetailContent.vue
@@ -1,0 +1,343 @@
+<template>
+  <div class="match-detail-content">
+    <div v-if="isJubileeGame" class="jubilee"></div>
+    <v-card tile class="pb-5 pt-1">
+      <v-card-title class="justify-center">
+        <v-row justify="space-around">
+          <v-col cols="1" class="pl-0 pr-0">
+            <v-card-subtitle class="pa-0 text-uppercase opacity-100">
+              <div>{{ $t(`gatewayNames.${gateWay}`) }}</div>
+              <div>{{ $t(`views_matchdetail.season`) }}: {{ season }}</div>
+            </v-card-subtitle>
+            <host-icon
+              v-if="match.serverInfo && match.serverInfo.provider"
+              :host="match.serverInfo"
+              style="padding-right: 0px"
+            />
+          </v-col>
+          <v-col v-if="!matchIsFFA" cols="4" align-self="center">
+            <team-match-info
+              :big-race-icon="true"
+              :left="true"
+              :team="match.teams[0]"
+            />
+          </v-col>
+          <v-col cols="1" class="text-center" align-self="center">
+            <span v-if="!matchIsFFA">{{ $t(`views_matchdetail.vs`) }}</span>
+          </v-col>
+          <v-col v-if="!matchIsFFA" cols="4" align-self="center">
+            <team-match-info :big-race-icon="true" :team="match.teams[1]" />
+          </v-col>
+          <v-col v-if="matchIsFFA" cols="6" align-self="center">
+            <team-match-info
+              class="ma-1"
+              :big-race-icon="true"
+              :team="match.teams[0]"
+            />
+            <team-match-info
+              class="ma-1"
+              :big-race-icon="true"
+              :team="match.teams[1]"
+            />
+            <team-match-info
+              class="ma-1"
+              :big-race-icon="true"
+              :team="match.teams[2]"
+            />
+            <team-match-info
+              class="ma-1"
+              :big-race-icon="true"
+              :team="match.teams[3]"
+            />
+          </v-col>
+          <v-col cols="1" />
+          <div v-if="showReplayDownload" class="subicon">
+            <download-replay-icon :gameId="matchId" />
+          </div>
+        </v-row>
+      </v-card-title>
+      <div class="pb-2">
+        <v-card-title v-if="isJubileeGame" class="d-flex justify-center">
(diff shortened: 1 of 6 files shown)


francislavoie:
Hmmm. I understand the motivation, but I think so many people are used to clicking on the row to copy the match link (especially for reporting a match etc) that it would break people's muscle memory and be annoying more than useful. Most people have a back button on their mouse so navigating back to the list after clicking is fast too.

francislavoie:
I think an option that wouldn't break semantics as much is a caret arrow to expand and still make the row clickable to go to the full page. But I'm still not convinced how much of a value add this is, having two ways to do the same thing doesn't seem like a good idea.

drakonli:
> I think an option that wouldn't break semantics as much is a caret arrow to expand and still make the row clickable to go to the full page. But I'm still not convinced how much of a value add this is, having two ways to do the same thing doesn't seem like a good idea.

- My motivation is this: when looking through player history I normally look at bunch of games as I know do others as well. Now, its quite inconvenient "leaving" the page like that and then returning, but also its easy to lose track of which specific game you were looking at (pressing back doesn't send you to that game position) and there's no way to check without reopening the page again. To me it seems accordion should have been the default behaviour here because it allows you to more effectively traverse multiple games without leaving the page

- I did make a new button to open the page next to "replay"(just in case, I attached a screenshot), so a person can still look at the page/report/etc. I understand your concerns about muscle memory, but I think the button is quite visible and understandable such that it wont create problems 🙂

gustav87:
How about having a ctrl+click option to open the match detail in a new tab?

```
              <v-row
                v-if="!isFfa(item.gameMode)"
                :class="{ 'cursor-pointer': !unfinished }"
                class="force-no-wrap"
                @click.exact="goToMatchDetailPage(item)"
                @click.ctrl="openMatchDetailPageInNewTab(item)"
              >
```

Reference: https://vuejs.org/guide/essentials/event-handling.html

drakonli:
> How about having a ctrl+click option to open the match detail in a new tab?
> 
> 
> 
> ```
> 
>               <v-row
> 
>                 v-if="!isFfa(item.gameMode)"
> 
>                 :class="{ 'cursor-pointer': !unfinished }"
> 
>                 class="force-no-wrap"
> 
>                 @click.exact="goToMatchDetailPage(item)"
> 
>                 @click.ctrl="openMatchDetailPageInNewTab(item)"
> 
>               >
> 
> ```
> 
> 

Sure, Its not a problem, but how would people know to do it? I felt like a separate button with a popular icon would communicate the purpose clearly

gustav87:
> > How about having a ctrl+click option to open the match detail in a new tab?
> > ```
> > 
> >               <v-row
> > 
> >                 v-if="!isFfa(item.gameMode)"
> > 
> >                 :class="{ 'cursor-pointer': !unfinished }"
> > 
> >                 class="force-no-wrap"
> > 
> >                 @click.exact="goToMatchDetailPage(item)"
> > 
> >                 @click.ctrl="openMatchDetailPageInNewTab(item)"
> > 
> >               >
> > ```
> 
> Sure, Its not a problem, but how would people know to do it? I felt like a separate button with a popular icon would communicate the purpose clearly

It's standard on the web to open things in a new tab with ctrl click.

gustav87:
`src\components\matches\PlayerMatchInfo.vue` has logic for opening a profile in a new tab if you want to check it out.

```
    function openProfileInNewTab() {
      if (!showPlayerInfo.value) return;

      const path = getProfileUrl(props.player.battleTag);
      window.open(path, "_blank");
    }
```

drakonli:
> `src\components\matches\PlayerMatchInfo.vue` has logic for opening a profile in a new tab if you want to check it out.
> 
> ```
>     function openProfileInNewTab() {
>       if (!showPlayerInfo.value) return;
> 
>       const path = getProfileUrl(props.player.battleTag);
>       window.open(path, "_blank");
>     }
> ```

I've added the ctrl+click ability on a row to open in new tab as an additional option now!

Though now that I think about it, did you mean this INSTEAD of the accordion or together?

gustav87:
Yeah, instead of the accordion

drakonli:
I really want this change for user experience, have been frustrated by this for years now. 

I assume you dont have like a community vote on UX issues?

If not, then I think we can just close this PR and maybe give me a list of issues that you guys approve and I could work with

gustav87:
Perhaps the logic could be swapped around, so that the button expands the accordion?

Because now the flow is as follows:
- Normal click: expand accordion
- Ctrl click: route to match detail in a new tab
- Button click: route to match detail in the same tab

gustav87:
There are some minor things like font size differences between the match details in the accordion and the match details in a new window. Some texts are smaller.

And you use `:class="{ 'cursor-pointer': isAccordionCellActive() }"` in multiple places, is this to use a pointer for only the items with currently open accordions? I think `isAccordionCellActive` isn't doing what the name implies, because it's always true in the player matches tab, and never true in the matches view.

gustav87:
I like the accordion though. It's a nice idea!

gustav87:
Closing due to inactivity

PR #1040, opened 2026-04-28
Fix map filter on finished matches

CoffMakesThings:
This fixes the map filter being broken for finished matches.

For seasons 10+, we have the MapName in the MatchUp collection.

In this PR we get the distinct MapNames for the season, offer them as dropdown options, and filter by them.

For seasons 1-10, we will stop offering the Map filter for now. Steps could be taken to make those work too, in a future PR.

This would benefit from an index on MatchUp.MapName.

image

[website-backend pr](https://github.com/w3champions/website-backend/pull/513)

### DIFF
--- a/src/store/match/store.ts
+++ b/src/store/match/store.ts
@@ -15,6 +15,7 @@ export const useMatchStore = defineStore("match", {
     allOngoingMatches: [] as Match[],
     matchDetail: {} as MatchDetail,
     mapNames: [],
+    mapNamesCache: {},
     status: MatchStatus.onGoing,
     gameMode: EGameMode.GM_1ON1,
     map: "Overall",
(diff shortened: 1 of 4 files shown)


CoffMakesThings:
Been working on this with @MichaelDeitz

src/store/match/store.ts
```diff
@@ -82,15 +90,27 @@ export const useMatchStore = defineStore("match", {
       this.SET_MATCH_DETAIL(response);
       this.SET_LOADING_MATCH_DETAIL(false);
     },
+    async loadMapNames() {
```
gustav87, line 94: You might want to cache the mapnames in the frontend because if you've fetched them once they are very likely to be the same the next time you need to fetch them.

CoffMakesThings: I'll try to add that.

CoffMakesThings: map names are now cached until the user refreshes the website

gustav87 (approved):

gustav87:
Rebase this one and we can merge it!

PR #1044, opened 2026-05-02
Add "Overall" column for map winrates stats

francislavoie:
In Grubby discord, Solstice suggested an Overall column for these stats to show winrates per map per race totalling all matchups. It's useful to give a sense of which maps are good to veto or not, regardless of specific matchup.

image

### DIFF
--- a/src/components/player/PlayerStatsRaceVersusRaceOnMapTableCell.vue
+++ b/src/components/player/PlayerStatsRaceVersusRaceOnMapTableCell.vue
@@ -6,11 +6,11 @@
       </td>
     </template>
     <div>
-      <span class="number-text w3-won">{{ stats.wins }}W</span>
+      <span class="number-text w3-won">{{ formatTooltipNumber(stats.wins) }}W</span>
       -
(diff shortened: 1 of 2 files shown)


src/components/player/PlayerStatsRaceVersusRaceOnMapTableCell.vue
```diff
@@ -44,8 +44,15 @@ export default defineComponent({
...
       default: undefined,
     },
+    ignoreCompareRace: {
+      type: Boolean,
+      required: false,
+      default: false,
+    },
   },
   setup(props) {
+    const formatTooltipNumber = (value: number): string => Intl.NumberFormat().format(value);
```
gustav87, line 54: didn't know about this function, nice!

gustav87 (approved):

src/components/player/PlayerStatsRaceVersusRaceOnMapTableCell.vue
francislavoie: Yeah this is to add thousands commas to the numbers, easier to read
image

PR #1045, opened 2026-05-05
feat: add head-to-head history to match detail page

MarkWind85:
Hey! This adds a head-to-head section to the 1v1 match detail page that shows the rivalry between two players — score, recent form, previous matches, map stats, and game duration patterns. Currently it only shows up when they've played each other more than once in a season — do you want it always visible regardless, or is the >1 game threshold the right call?

It calls `MatchService.retrievePlayerMatches()` (`GET /api/matches/search`) so everything runs through the existing API. I followed the existing patterns for components, styling, and i18n as closely as I could.

You can see it in action on any 1v1 rematch — `/match/69f96aab6e63ce402f26ed91` (JandeX vs Hera, S24) is a good example.

This is my first contribution so happy to hear if anything needs adjusting!

~Mark (thehighnotes)


francislavoie:
Can you show some screenshots of what this looks like?

MarkWind85:
Sure! this is the overview - image

This is with the season list (without querying the API for all seasons, which i think is expensive/inefficient, i thought to keep it in, but i can also reduce the number of seasons shown to last #)
image

francislavoie:
Cool!

The L-L-L-W thing should look more like this (maybe reuse the same component)

image

The seasons picker, do we have an API to fetch which seasons there was any h2h data? Showing all seasons seems not right probably, and either way the season picker should have a max height to not run so long.

The durations in the table should be like this, with the bar below, for consistency

image

It should probably repeat the usernames in the h2h section to further clarify who is left and who is right when it comes to the win-loss record (who counts as red and who as green? can be misleading since the top uses red/green for results of _this match_).

MarkWind85:
Ha - i did actually have the usernames in the H2H section but then removed them. Alright good feedback - I'll be sure to make those changes. And yeah the season picker - i don't there is a specific one for that, i'll have another look in the swagger, but i want to not hard-query the match history over seasons for each call :D in order to know which seasons to show and which not.. 

I'll fix the seasons by limiting the options to the last few seasons.. (which makes the most sense probably?). As well as a max height (in case an api later becomes available).

I'll work on another commit, thanks!

francislavoie:
I think it would be fine load current + previous season on initial load, helps give a bit more historical data especially in early days of a season.

If season is an optional parameter in the API, while still having a max results, could just fetch last like 20 or 50 games across seasons and show it that way.

MarkWind85:
Matches api says season is optional, but doesnt behave that way (returns 0). For now i'll work with the current + previous and a limited season picker. Later iterations can improve this :)

MarkWind85:
Ive processed the feedback :)  - I think this works pretty well now. Let me know if you have any remarks. 

Here';s the screenshots (this feature is extra cool for known high-level matchups) 

picker
s24-s23-all

francislavoie:
The "days ago" time should be to the left of the duration, with more padding in between (it's basically touching eachother, too close), to match the match history lists elsewhere.

The total width the whole section can be shrunk as well, lots of dead space in the center below the Recent Performance thing.

Also I'm not sure what the head-to-head API returns, it may return games where the left player uses a different race than the current game as well, so that should be indicated too most likely, especially if they play Random for example.

We don't tend to use "W" and "L", I think the map name should be colored green/red instead most likely, in addition to having both races as info.

MarkWind85:
You're right about the races, it does return all races played. - I implemented your other feedback too, let me know if anything needs changing (before i commit again). I kept it more in line now with the existing width of the match-breakdown. Ive not included the names again above the races - would this work you think? 

edit; show all indicates that there are matches hidden < 2 min. Can be defaulted to showing all instead

image

francislavoie:
Yeah looks a lot better. I don't think the map names need the underline though, as long as the row has a hover style (brighter bg or w/e, pointer cursor) to make it clear it's clickable. Underlined links aren't really used anywhere else on the site.

I think that's the last comment I have until I find a moment to run it locally and take it for a spin. Looks pretty good!

Edit: Oh yeah, don't hide any short matches, that's not consistent with anywhere else.

MarkWind85:
Brilliant idea; didnt include that yet; match links :)!  Alright ill remove the filter for < 2 min and will remove the underlines.  New commit coming in

francislavoie:
Also hmm, I think the "patterns" thing is not that useful, just game lengths isn't really that interesting. I think we can just remove that part, to save a bit of space. The "By Map" info is also probably not that interesting either. I think that kind of info should be left to "Statistics" type pages, we have a Statistics tab on a player's profile. Map-specific results against a specific player is generally less interesting than overall performance on a certain map vs a certain race, which is already covered.

MarkWind85:
Given your feedback, the page will be much more focussed, which makes me wanna remove the season picker and introduce lazy loading for the previous seasons on a per season basis. And if two consecutive seasons offer no matches for this h2h, then lazy loading is disabled for any more seasons. - will commit this and remove the picker

MarkWind85:
For consistency and clarity I made sure that the current match being viewed is highlighted..  This has some limitations, since cache resets upon clicking on a new match ; a user would have to recall the API. 

Ive got 5 or 7 more features lined up, one of which i think may be that navigation behavior for match clicks: UX could be improved with maintaining some cache instead of using window.location.href for matchups.

image

francislavoie:
Are you setting a limit of amount of games? I think if we do it that way the recent games should be in a scrollable box, possibly with a "Load More" button at the bottom. I think it can stick to just loading current and prev season to avoid too many API calls and load more to trigger fetching more.

MarkWind85:
I was planning on doing it per season..   instead of 'load more'. 

Edit; i can set a limit with an in-season load more, in case that ever happens.
And yes ill box it :) - good idea

francislavoie:
By season is fine too, but I think it should read as "Load more" and just fetch the next season.

I like highlighting the currently shown match, that's good. Make sure that looks okay on all themes btw.

You said cache resets? Surely that can be adjusted to keep the loaded head-to-head when going from one game of the same player to the next.

MarkWind85:
Already checked themes, its consistent with theming :) same for the hover action

and cache;  it's because MatchDetail.vue doesn't watch for route param changes, so we have to use a full page reload (window.location.href) to navigate between matches. That destroys all client state including the H2H cache.    

It can be persisted in session storage, but really the proper navigation would be the more elegant solution..

francislavoie:
Yeah we should fix the routing then, that's good to change.

MarkWind85:
Happy to do it sometime this week :) [edit: already made the PR] as well as some mobile scaling here. Those are the two more technical features im keen on tackling

MarkWind85:
Committing 

image

francislavoie:
Alright I checked out the branch and played around with it, bunch of little feedback details:

- Clicking another match uses `window.location.href` to navigate which means it's a full page reload. That shouldn't happen, it should use the router to navigate to another match with `router.push({ path: `/match/${matchId}` });`, but I think this requires https://github.com/w3champions/website/pull/1046 to actually work, of course. (I tried it out by cherrypicking and it worked fine, so that's good).
- The scroll box has top padding which leaves a small gap above the first season header which looks not good. The top padding should be removed.
- The season headers have a bright 1px bottom border, I don't think this looks good, we should remove that border I think.
- The little left border on the highlighted current match causes the content to be shifted by a couple pixels to the right which misaligns it. The fix would be to also apply a border for all matches but make it transparent, so they're aligned perfectly.

MarkWind85:
Awesome: Yes with https://github.com/w3champions/website/pull/1046 merged I can set this PR to the route navigation.  
- Ah on my end i had the gap dissapear.. I'll do some more digging and make sure it works consistently.
- - alright ill remove the border
- - Oh youre right, will do that. 

New commit coming with these points

MarkWind85:
Screenshot: Alignment is now fixed - season Border is removed along with the other improvements mentioned in commit. 

image

src/views/MatchDetail.vue
```diff
@@ -113,6 +113,15 @@
               />
             </v-col>
           </v-row>
+          <match-head-to-head
+            v-if="isCompleteGame && !matchIsFFA && playerBattleTag && opponentBattleTag"
```
francislavoie: This condition isn't sufficient, it still shows head-to-head for 4v4 games which doesn't make sense. It should only show if there's only two players total on separate teams I think.

MarkWind85: Agreed!  so ill do: Only 2 teams, only 2 players. which makes it one-on-one mode-agnostic - so 1on1 tournament games for instance are included

francislavoie: I'll fix this, working on a followup commit cause I noticed some other tangential issues that I'll add on top of this.

Edit: nvm lol I'll rebase

francislavoie:
Pushed a commit to fix match results in situations where the battletag in the score data doesn't match (sometimes the game reports it using the wrong battletag, very odd). For example https://w3champions.com/match/6973fb77644e62fcdc7785ca would currently fail to render properly, see console logs for a slew of errors.

MarkWind85:
@francislavoie Sounds similar to what i hoped https://github.com/w3champions/website/pull/1046 would prevent too "Includes a guard against stale API responses — if a user navigates between matches quickly on a slow connection, out-of-order responses are discarded to prevent showing the wrong match data. Uses the immediate: true pattern, matching the existing approach in Player.vue."

francislavoie:
Not related. The data from the backend is actually "bad" so we need to do some gymnastics to fix it. We could do a fix in the backend eventually too, but this is easier.

francislavoie:
It's deployed now btw! But found some issues

https://w3champions.com/match/69fbb53b6e63ce402f27c5ec

If you click Load More a whole bunch, you'll see it says "Last 50 games" but there's more than 50 (44-39 is more), and only a subset of the games actually show up in the scrollbox.

MarkWind85:
Yup replicated it, looking into it now

PR #1050, opened 2026-05-08
fix: H2H load more was silently dropping new matches

MarkWind85:
The 15-match cap in MatchHeadToHead was leftover from an early
exploration where "load more" was going to fetch in chunks of 15
matches at a time. We ended up going per-season instead and I
forgot to remove the cap, so loaded matches got appended to the
data but sliced out before rendering — count went up, rows didn't.

Reproduced on https://w3champions.com/match/69fbb53b6e63ce402f27c5ec
and confirmed fixed.

~Mark (thehighnotes)

### DIFF
--- a/src/components/match-details/MatchHeadToHead.vue
+++ b/src/components/match-details/MatchHeadToHead.vue
@@ -146,6 +146,10 @@ export default defineComponent({
       type: Number as PropType<Gateways>,
       required: true,
     },
+    gameMode: {
+      type: Number as PropType<EGameMode>,
+      required: true,
+    },
   },
   setup(props) {
     const router = useRouter();
@@ -231,7 +235,7 @@ export default defineComponent({
         0,
         props.playerBattleTag,
         props.opponentBattleTag,
-        EGameMode.GM_1ON1,
+        props.gameMode,
         ERaceEnum.TOTAL,
         ERaceEnum.TOTAL,
         props.gateway,
--- a/src/components/match-details/MatchHeadToHead.vue
+++ b/src/components/match-details/MatchHeadToHead.vue
@@ -1,5 +1,5 @@
 <template>
-  <div v-if="!matchIsFFA && hasData" class="mt-8 mb-4">
+  <div v-if="hasData" class="mt-8 mb-4">
     <v-card-title class="d-flex justify-center">
       {{ $t("components_matchdetails_matchheadtohead.headToHead") }}
     </v-card-title>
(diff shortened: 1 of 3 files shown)


francislavoie:
It still says "Last 50 games" tho regardless of having more than 50 loaded

francislavoie:
Oh weird, the H2H results also include results from other modes than the current one. Doesn't it have a mode filter for the API? We should use that to only get results from matching modes. See http://localhost:8081/match/69fa6cc76e63ce402f274b1b the last results are FFA matches e.g. http://localhost:8081/match/66258c4768c045f85405e91b which should not show up when looking at a 1v1 match

MarkWind85:
It does have a filter..  Let me check that. The count to 50 cap is removed btw. But looking into the filter reliability now

MarkWind85:
so the filter i had applied, worked on page load but no on 'load more'. 
the API unfortunately doesnt allow filters in the way i want to show the matchup history.. It does support gamemode, but you'd lose tournament or Legion or any other future 1v1 type. because gamemode only accepts 1 mode, not a list. 

So, the thing i did was render filter it (client side). But i didnt set it on subsequent loads. Apologies :). Am doing that now and testing it with the cases you found :)

francislavoie:
I think it's correct to only show melee 1v1 and excluding legion TD etc. I don't think mixing head-to-head between totally separate modes makes sense. Also tournament is entirely inactive right now, and I would probably want to see only head to head in tournaments separate from regular matchmaking.

src/components/match-details/MatchHeadToHead.vue
```diff
@@ -238,7 +231,7 @@ export default defineComponent({
         0,
         props.playerBattleTag,
         props.opponentBattleTag,
-        EGameMode.UNDEFINED,
+        EGameMode.GM_1ON1,
```
francislavoie: Well, there's other kinds of 1v1 modes (warhammer 1v1, ATR 1v1, etc) so hard-coding to specifically GM_1ON1 isn't right, it should inherit the mode from the mode of the currently loaded match.

MarkWind85: THat's solid :) do you think the UX would speak for itself or should the H2H History indicate the mode (for the user i mean)?

francislavoie: I think the UX is clear enough for now.

But I'm thinking we might shift it later to look just like a the regular match history view, with usernames (with colors) on either side and map+mode etc. But I think it's fine as-is. I think we'd only do that if a bunch of users complain that they don't understand it. I think it's fine.

PR #1057, opened 2026-05-08
Enhances UI with loading states and dynamic titles

francislavoie:
### Summary
Improves the user experience by introducing loading spinners for matches and match details, standardizes the display of gateway selection, and dynamically sets page titles for player profiles.

### Key Changes
*   **Adds Loading Spinners**: Implements loading indicators for the match list view and individual match detail pages, providing visual feedback during data fetching. The match list also includes logic to prevent stale data from older requests.
*   **Standardizes Gateway Visibility**: Introduces a shared constant and helper function (`isGatewayNeededForSeason`) to consistently determine when gateway selection is relevant (e.g., for older seasons) across multiple views (MMR Distribution, Country Rankings, Player Profile, Match Details, Rankings). The gateway select component is now hidden when not needed.
*   **Dynamic Player Profile Titles**: Updates the browser tab title for player profile pages to include the player's battle tag and the current tab name (e.g., "PlayerName - Profile"), enhancing navigation context.

### Resolved Issues
*   **Fixes Profile Data Flash**: Resolves an issue where navigating to a new player profile would briefly display the previously loaded profile's data.
*   **Minor Style Adjustments**: Adjusts padding in the Head-to-Head match list and adds margin to duration cells for better visual alignment.

### Dependency Updates
*   Updates `package-lock.json` via `npm i`, shouldn't happen again now that @gustav87 updated his `npm` to latest version.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -281,6 +281,27 @@
         "win32"
       ]
     },
+    "node_modules/@emnapi/core": {
+      "version": "1.10.0",
+      "resolved": "https://registry.npmjs.org/@emnapi/core/-/core-1.10.0.tgz",
+      "integrity": "sha512-yq6OkJ4p82CAfPl0u9mQebQHKPJkY7WrIuk205cTYnYe+k2Z8YBh11FrbRG/H6ihirqcacOgl2BIO8oyMQLeXw==",
+      "license": "MIT",
+      "optional": true,
+      "dependencies": {
+        "@emnapi/wasi-threads": "1.2.1",
+        "tslib": "^2.4.0"
+      }
+    },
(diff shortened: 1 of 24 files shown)


francislavoie:
Graph improvements:
- MMR graph's purple bars changed to markers, easier to read and avoids needing the explanation text beside the graph.
- MMR graph season changing is smoothed out by putting the loading spinner on top so the data animates when the next season is loaded in
- Games per Day graph now has hover dimming, ported from the lag-report charts that I set up recently - hovering on a game mode line dims the other ones so you can see it on its own, also dims the color in the legend so you can see where's in the legend the one you're mousing over.

PR #1059, opened 2026-05-09
Dark-mode dropdown affordance + matches status toggle

MarkWind85:
Sitewide dark-mode fix: transparent dropdown buttons looked like plain
text since the transparent bg strips Vuetify's elevation. Added a
`w3-dropdown-button` class and applied it across all Select components
plus the Rankings/CountryRankings menus. Floating menus had the same
problem on dark backgrounds, so `.v-overlay__content` gets a subtle
border and faint inverted shadow. Makes dark themes feel visually
consistent with the lighter ones.

Also turned the matches status filter into a single-button toggle since
it only ever had two options (Now Playing / Finished) and they're
unlikely to grow. Note: the
`components_matches_matchesstatusselect.selectstatus` i18n key is now
unused but I left it since I understand locales are managed via Google
Sheets.

### DIFF
--- a/src/components/common/DurationSelect.vue
+++ b/src/components/common/DurationSelect.vue
@@ -7,6 +7,7 @@
     <template v-slot:activator="{ props }">
       <v-btn
         tile
+        class="w3-dropdown-button"
         style="background-color: transparent; min-width: 180px;"
         v-bind="props"
       >
--- a/src/components/common/GameModeSelect.vue
+++ b/src/components/common/GameModeSelect.vue
@@ -1,7 +1,7 @@
 <template>
   <v-menu location="right" transition="fade-transition">
     <template v-slot:activator="{ props }">
-      <v-btn tile style="background-color: transparent" v-bind="props">
+      <v-btn tile class="w3-dropdown-button" style="background-color: transparent" v-bind="props">
         <v-icon size="x-large" start>{{ mdiControllerClassic }}</v-icon>
         {{ gameModeName() }}
       </v-btn>
--- a/src/components/common/GatewaySelect.vue
+++ b/src/components/common/GatewaySelect.vue
@@ -1,7 +1,7 @@
 <template>
   <v-menu location="right">
     <template v-slot:activator="{ props }">
-      <v-btn tile style="background-color: transparent" v-bind="props">
+      <v-btn tile class="w3-dropdown-button" style="background-color: transparent" v-bind="props">
         <v-icon size="x-large" style="margin-right: 5px">{{ mdiEarth }}</v-icon>
         {{ $t(gateway.name) }}
       </v-btn>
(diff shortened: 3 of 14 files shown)


francislavoie:
Can you show screenshots with a comparison?

MarkWind85:
The current:
image

The new:
image

It doesnt change lightmode, from my eenvironment:
image


(ignore the width variance, its my browser-zoomlevel)

francislavoie:
I see. I think it doesn't look quite right yet, it just looks like a bright white line to me and not so much like a glow 🤔 it is a tricky thing to fix visually

I think what we should do instead is make the buttons bg slightly gold (dark gold bg with bright gold text) and then the regular shadows will look okay. Like maybe take the current hover color and make that the regular BG, then make the hover a bit brighter than that.

MarkWind85:
Yeah.. you are totally on point; i actually experimented with that too (or at least with the idea of changing button bg colors), and it does look better.. But i didnt want to go too far with the style-change. 

I'll get on it and share some screenshots to verify

MarkWind85:
So buttons are universally set to transparant background; either i change that fundamental css property (and then do a proper site-wide check on all button elements and make necessary fixes in a new css standard.. or.. use !important for now to the new css class. 

Are you ok with !important for now?

francislavoie:
Yeah... vuetify overrides are annoying, we do have plenty of `!important` already for theming unfortunately, so I think it's fine.

MarkWind85:
So the regular shadows dont show: 
image

So i added a gold shadow..  i am no visual style master ha so please let me know what you think 

image

The hover effect is now a little brighter for contrast and visual feedback :)

francislavoie:
Hmm. Yeah I guess that isn't too nice either cause it breaks the affordance of the hover styles used elsewhere (nav buttons etc). I think that disqualifies that idea now that I see it implemented.

I don't really know how's best to solve it. It's awkward.

I think we can go back to the original idea, but make it even more subtle (was too bright, maybe use a grey instead of transparent white -- actually the right side looked fine in the first try but the bottoms looked too bright). I think the vertical dividers between the buttons should be consistent with the bottom somehow, it looks odd when there's more bottom shadow than between or w/e.

I think Show Heroes should use the same styles.

MarkWind85:
Yeah i tried to adhere to the shadow pattern that the light themes have on the buttons, which also have more shadow on the bottom.

MarkWind85:
image
what do you think  @francislavoie 
i'll commit if you have no further feedback :)

francislavoie:
I think that's good! Can you apply it to Show/Hide Heroes as well?

MarkWind85:
Oh of course; keep its position?
image

francislavoie:
Yeah leave it where it is.

PR #1067, opened 2026-05-16
Match detail: mobile-responsive header and hero stats

MarkWind85:
Mobile responsiveness for the match detail page. The header now stacks
season info, player names, and VS vertically on narrow viewports while
keeping the original mirrored layout on desktop. Race and flag icons
flip to the same side when stacked so both players read consistently.
The VS column loses padding on mobile so names can close in, then
restores it at md+.

The hero stats section was restructured — icons and values now live in
their own columns instead of being paired per-row inside
MatchHighlights. This fixes the alignment drift that happened when
different number widths (1 vs 621 vs 12) pushed icons to different
positions. Hero portraits scale up on mobile (cols 2 below md) with a
100px cap so they don't blow up on wider tablets. Spacer columns hide
on mobile to reclaim space, and all stat/icon padding scales through
three breakpoints.

MatchHighlights was only used in MatchDetailHeroRow so it's been
absorbed — the comparison logic now lives inline and the component is
deleted. Also fixed textClass in PlayerMatchInfo from a ref to a
computed so it reacts when the left prop changes dynamically on resize.
And corrected four `false: true` prop typos in MatchDetailHeroRow to
`required: false` (no runtime change, just says what it means).

I didn't touch the scaling behavior of the game-stats (units killed etc.). 

~Mark (changed account name from thehighnotes to real name btw :P - for work)


MarkWind85:
Pictures!
Current:
image

New: 
New-Mobile
New-Small-Mobile

And New Desktop / no scaling:
New-Desktop

francislavoie:
Cool, responsiveness is a good thing to improve 👍 

So actually we have much nicer match detail views in https://github.com/w3champions/twitch-extension that I worked on some time ago but never got around to porting to the website. I don't have screenshots for you right now but that's something to look into.

MarkWind85:
Oh! i'll have a look :)

MarkWind85:
I may delete this PR and resubmit the twitch version later on

MarkWind85:
Closing in favor of a more comprehensive match detail redesign that ports the twitch extension layout.

PR #1068, opened 2026-05-16
Redesign 1v1 match detail (port from twitch extension)

MarkWind85:
Brings the match detail layout from the twitch extension over to the
website for 1v1 matches, with responsive scaling added on top.

The header now shows player names large and bold with race icons around
VS, MMR on its own line with coloring and a crown for the winner.
Season info and the download button share a top bar. Country flags
stay.

The stats area is rebuilt as a 3-column grid — hero icons and per-minute
rates on the sides, all nine comparison rows in the center using a new
ScoreStat component with the same stat icons from the twitch extension.
This replaces both MatchDetailHeroRow and PlayerPerformanceOnMatch for
1v1s.

Below 850px the layout adapts: heroes move above the stats, per-minute
rates join the center grid as regular stat rows, and the side columns
drop out. Names truncate with ellipsis, hero icons scale down. On
narrow phones (Galaxy S20, iPhone SE, etc.) the container goes
edge-to-edge to reclaim space.

Map, duration, and start time sit below the stats matching the twitch
extension placement.

Team and FFA views are unchanged. MatchHighlights is deleted since
ScoreStat replaces it.

~Mark (aka thehighnotes)


MarkWind85:
Current:
OG-Mobile

New Desktop:
Match-New-Desktop

New Mobile:
mobile-Matches-New

marcoabreu:
Could you introduce the leagues and ranks as well please? Want to reduce the focus on pure MMR

MarkWind85:
Are you sure?  Isnt MMR still scene-wise the way skill-level is measured/talked about most of the time?

For LOL and SC2 ranks are dominant discourse, but i notice for Wc3 and AOE2 that MMR/ELO are..

marcoabreu:
Yes and it's a problem. We can't hide MMR, but I want to avoid putting it into the foreground or treating it higher than MMR.

MarkWind85:
I hear ya, alright :) !

francislavoie:
Looks great, thank you!

I'll do a more thorough review in a couple days when I'm less busy.

MarkWind85:
Alright now with Leagues & Ranks (clickable to the specific League/Div)
On mobile 
image

On Desktop:
image

Commit incoming

marcoabreu:
Looks nice!

src/views/MatchDetail.vue
```diff
@@ -11,55 +11,18 @@
...
-                  :team="match.teams[2]"
-                />
-                <team-match-info
-                  class="ma-1"
-                  :big-race-icon="true"
-                  :team="match.teams[3]"
-                />
-              </v-col>
-              <v-col cols="1" />
+          <div v-if="isOneVsOne && !loading" class="ovo-header">
```
francislavoie: I think we should break this down into subcomponents now for each type of mode since we'll need to have team mode and FFA versions as well.

src/views/MatchDetail.vue
```diff
@@ -244,17 +537,36 @@ export default defineComponent({
...
+    const matchDuration = computed<string>(() =>
+      formatSecondsToDuration(match.value.durationInSeconds),
+    );
+    const playedDate = computed<string>(() =>
+      formatTimestampStringToDateTime(match.value.startTime),
+    );
+    const ffaPlayers = computed<PlayerScore[]>(() => [
+      ffaWinner.value,
+      ...ffaLosers.value,
+    ]);
```
francislavoie: not too sure why these were reformatted, I think they were fine as oneliners.

src/views/MatchDetail.vue
```diff
@@ -278,11 +590,99 @@ export default defineComponent({
...
+      {return Math.floor(p.currentMmr - p.oldMmr);}
+      return 0;
+    });
+
+    const mmrChangePlayer2 = computed<number>(() => {
+      const p = match.value?.teams?.[1]?.players?.[0];
+      if (p?.oldMmr && p?.currentMmr)
+      {return Math.floor(p.currentMmr - p.oldMmr);}
+      return 0;
+    });
```
francislavoie: this assumes two players, so it would need to go inside the 1v1-specific subcomponent. For team modes it would need reconsideration anyway since there's many more players. Also the code here is identical between the two so it should be a generic function invoked for either player

src/views/MatchDetail.vue
```diff
@@ -278,11 +590,99 @@ export default defineComponent({
...
+      0: "Grand Master",
+      1: "Master",
+      2: "Adept",
+      3: "Diamond",
+      4: "Platinum",
+      5: "Gold",
+      6: "Silver",
+      7: "Bronze",
+      8: "Grass",
+    };
```
francislavoie: These should use the translation strings instead of hardcoding here. There's consts/enums somewhere else for this stuff anyway.

src/views/MatchDetail.vue
```diff
@@ -278,11 +590,99 @@ export default defineComponent({
...
+          match.value?.teams?.[0]?.players?.[0]?.ranking?.leagueOrder ?? -1
+        ] ?? "",
+    );
+
+    const leagueNamePlayer2 = computed(
+      () =>
+        leagueNames[
+          match.value?.teams?.[1]?.players?.[0]?.ranking?.leagueOrder ?? -1
+        ] ?? "",
+    );
```
francislavoie: in general the `player1`/`player2` properties being repeated logic is messy

src/views/MatchDetail.vue
```diff
@@ -278,11 +590,99 @@ export default defineComponent({
...
+      0: "255, 215, 0", // Grand Master — gold
+      1: "180, 130, 255", // Master — purple
+      2: "0, 200, 200", // Adept — teal
+      3: "100, 180, 255", // Diamond — blue
+      4: "180, 220, 255", // Platinum — ice blue
+      5: "255, 215, 0", // Gold — gold
+      6: "192, 192, 192", // Silver — silver
+      7: "205, 127, 50", // Bronze — copper
+      8: "100, 200, 100", // Grass — green
+    };
```
francislavoie: These should also be moved somewhere common. Are we doing this glow somewhere else already? Or this is a new invention?

src/views/MatchDetail.vue
```diff
@@ -278,11 +590,99 @@ export default defineComponent({
...
+
+    function leagueGlow(leagueOrder: number): string {
+      const color = leagueGlowColors[leagueOrder] ?? "255, 255, 255";
+      return `drop-shadow(0 0 5px rgba(${color}, 0.5))`;
+    }
```
francislavoie: I think I prefer this to use CSS variables, I don't like having CSS as "data" in view functions if it can be avoided.

src/views/MatchDetail.vue
```diff
@@ -278,11 +590,99 @@ export default defineComponent({
...
+      router.push({ path: getProfileUrl(battleTag) });
+    }
+
     const playerScores = computed<PlayerScore[]>(() => {
       const { playerScores, match } = matchStore.matchDetail;
       if (matchIsFFA.value) {
         const ffaMappedPlayerScores = playerScores.map((playerScore) => {
-          const battleTag = match.serverInfo.playerServerInfos[playerScore.teamIndex].battleTag;
+          const battleTag =
+            match.serverInfo.playerServerInfos[playerScore.teamIndex].battleTag;
```
francislavoie: I'm also not sure why these line wrappings happened. Are you using prettier in your IDE on autosave or something? We use dprint on this project which doesn't do this.

francislavoie (changes requested):

src/views/MatchDetail.vue
MarkWind85: Agreed!

src/views/MatchDetail.vue
MarkWind85: Yeah same as first, this will be tackled by generalizing it :)

src/views/MatchDetail.vue
MarkWind85: It was a new invention; i wanted to pronounce the league Icons a bit more specifically for this page given Marco's feedback.

src/views/MatchDetail.vue
MarkWind85: Yeah agreed. Will keep that in mind for future work :)

src/views/MatchDetail.vue
MarkWind85: is there?  League names are Hardcoded in 3 separate places — PlayerLeague.vue (all lowercase), PlayerMatchInfo.vue, and our MatchDetail.vue. There's no centralized enum or i18n keys for bare league names - i think there should be though.

So i think creating an actual shared constant is the next step.

src/views/MatchDetail.vue
MarkWind85: Ah that was generated - i'll be sure to review on dprint before committing.

src/views/MatchDetail.vue
MarkWind85: Will implement it, and apply to the mentioned files.

MarkWind85:
@francislavoie Ive applied the edits. I noticed that there was no standard for Grandmaster vs Grand Master - so i settled on the server-side info; Grand Master

src/components/match-details/MatchDetailOneVsOne.vue
```diff
@@ -0,0 +1,470 @@
...
+            />
+            <span class="ovo-league-name">{{ leagueName(1) }}</span>
+            <span v-if="player(1).ranking?.division" class="ovo-mmr-label">
+              {{ player(1).ranking?.division }}
+            </span>
+            <span v-if="player(1).ranking?.rank" class="ovo-rank-num">
+              #{{ player(1).ranking?.rank }}
+            </span>
+          </router-link>
+          <span v-if="player(1).won" class="ovo-crown">👑</span>
```
francislavoie: I think we want to use a proper icon instead of emoji for this. In the launcher we have a fist icon (I'll fetch it for you), we have another crown we use in a few spots (the w3c logo itself). I'm not sure which one we should use though, it's a bit up for debate.

@Kaenshin which would you want us to use?

francislavoie: Image

Kaenshin: I think the fist is better :D

Kaenshin:
image
Here is some small feedback if you are willing to work on it more and polish.

francislavoie:
@Kaenshin 
- Map thumbnail is not possible yet (we don't have the map images available for web use, launcher shows it via pulling it from the map files)
- I think it would be weird to put the "diamond" icon after the #rank, though I would generally agree with mirroring icon position. It's a tricky one because of the information hierarchy for league/division/rank
- Server icon I'm not sure about because it depends where we'd put it for other versions of this screen (team modes, FFA). We might not have a nice way of putting it in the center like that for those, so it might be strange. So I think leaving it top-left is probably fine, at least until we've done the other versions as well then we can consider them together as a whole.

I agree with the rest :+1:

francislavoie:
@Kaenshin how do you think we should lay out the duration/time? The duration bar kinda assumes right (usually) or left alignment cause it needs to grow from some origin point to have some visual meaning, but in this layout we have that info centered so the origin point is the middle which is kinda weird.

Re H2H history, I think we could just drop the `History` word.

MarkWind85:
@francislavoie  i think the mirroring would mean: 

League Icon - League Name - Div - Rank - Winning Fist  [spaces]  Rank - Div - League Name - League Icon 

Could work - its a typical format that audiences know i think (at least I do)

francislavoie:
I don't think we should mirror the rank stuff. It would just be confusing.

Kaenshin:
> @francislavoie i think the mirroring would mean:
> 
> League Icon - League Name - Div - Rank - Winning Fist [spaces] Rank - Div - League Name - League Icon
> 
> Could work - its a typical format that audiences know i think (at least I do)

image
You can look how on this mockup it works, maybe this will inspire you to some solution, but eighter way it's good :D

Kaenshin:
> @Kaenshin how do you think we should lay out the duration/time? The duration bar kinda assumes right (usually) or left alignment cause it needs to grow from some origin point to have some visual meaning, but in this layout we have that info centered so the origin point is the middle which is kinda weird.
> 
> Re H2H history, I think we could just drop the `History` word.



> @Kaenshin how do you think we should lay out the duration/time? The duration bar kinda assumes right (usually) or left alignment cause it needs to grow from some origin point to have some visual meaning, but in this layout we have that info centered so the origin point is the middle which is kinda weird.
> 
> Re H2H history, I think we could just drop the `History` word.

image
Maybe something like this? Although I agree that maybe bar is overkill, it works when we have something to compare to in this case we don't

MarkWind85:
@Kaenshin @francislavoie  
What decisions are we settling on? or is there an in-between choice we can make as we figure out the finer details for a later PR? Happy to process the direction you guys want it to go in

francislavoie:
I think every point was discussed to agreement already, just needs finishing up.

MarkWind85:
ah! I totally missed that, thought the style discussion was ongoing. Cheers i'll finish up with the feedback mentioned. Sorry for the misread there

MarkWind85:
@Kaenshin ; i love your design concept. But i think it deserves a website-wider discussion to see if we can enhance the styling.. As I think it works because of the whole style :).  So i kept the style a little as is, but would love to explore more in a new PR?

Mobile
MobileMatch

Desktop
DesktopMatch

francislavoie:
Can you show an example with a real game (heroes etc), hard to really review fairly with a bad dataset 😅

Why are the division names hidden on mobile? 🤔

The division icon should be the same width as the race icon so that the text beside it aligns with the player names vertically. See the right-side player, "Master" is not aligned vertically.

I think the `#5` text should be grey? Not the division number. Like "Diamond 3" in white then `#5` grey after. Having the division number grey disconnects it from the division name which is awkward

The heroes on mobile seem to leave an empty gap if there's none, I think it should reduce to nothing if there's no heroes for either player (can happen with custom game modes which are 1v1 but have no heroes reported)

I think duration should go below start, can remove "Time" from "Start Time" (it's obvious it's a date).

Make the left player gold or white in H2H to make it clear who is considered the target/main player of the two (who the W belongs to). One of Kenshin's feedback points.

I was thinking we should add tooltips on most places that show start time of a game to show the "X minutes ago" type text, in match history views etc. Also H2H should standardize to use that same style (extract it to a "start time value" component with the date + tooltip?)

MarkWind85:
So i thought to hide division names on mobile, as it creates a ui layout challenge to keep everything together. If you have a good idea on how to solve that neatly i'm all ears. 
Keep in mind - the mobile layout as shown isnt shown with mobile viewport-width, just the layout, the mobile viewport width is actually much smaller-  which is why i made the decision to make it icons only. I kinda figured that full text to Icon is a standard practice in design when scaling to mobiles.

So i made League consciously bigger, to communicate its emphasis while leaning less on the MMR (inspired by Marco's request). Do you still want me to make a change to make it equal to Race icon? Same for Rank number, again this was influenced by Marco  mention of league and rank.  
@francislavoie If you can confirm one way or another i'll make the changes if needed.

Good call on no heroes testing; i actually didnt look at that - will make those tweaks for the empty space on no heroes.

And yes, sorry, i missed the feedback kaenshin gave on the winner emphasis, i did mean to include that change..

Regarding tooltips: would that be cleaner as a new PR? As i'd have some functional design questions to scope it correctly :).

I'll make the tweaks mentioned and let me know what you want with regards league-icon/rank/div/.

---- 
Screenshots below are just for refrence with PRD data; edits arent yet included

Here's a mobile viewport example;
image

The the other pictures with prd examples
image

image

MarkWind85:
Double check before i commit @francislavoie 

I couldnt maintain a Gold winner-color as it becomes illegible on light themes. So felt like i had to resort to theme default. 

---
Light with heroes:
desktop-light
mobile-light

---
Dark with heroes:
desktop-w-heroes
mobile-w-heroes

---
Dark without heroes:
desktop
mobilematch

francislavoie:
I think the problem is the font and icon sizes stay too big on mobile, it should reduce the sizes when going below the mobile breakpoint to let everything fit cleanly.

Also the icon sizes aren't consistent, the fist is bigger than races on mobile but is (correctly) the same size as the races on desktop. Also the league icon should match those sizes as well, but it's not consistently following them.

Could consider wrapping the country and race icons onto the next line on mobile to give more room for the usernames (two rows of content instead of just one, on either side of the VS)

francislavoie:
Another minor thing: the original design had the `SEASON #` in gold, now it's white for some reason, it should stay gold. Also remove the `:` colon, looks better without I think. We could also be using the season icons instead of just the plain number (see top of profile page for those)

MarkWind85:
I see the profile pages using both; left has the icon, right has the text.. I think the icon itself is clear enough right? i'll use that and see what we think

francislavoie:
On the launcher I made the dropdown picker for season on match history use the icons, I just never got around to applying it to the season pickers on the website, but I think it's a nice visual pop to use them where possible (and also tooltips to ensure there's an escape hatch is people find it hard to read)

MarkWind85:
yeah, will make sure to include the tooltip

MarkWind85:
hmm this feels a little weird doesnt it?  (the newline for the flag and race icon).
I mean i could make it sit along with the league + rank line.. but that feels weird too..  What do you think? Keep it like this? 

(just an in between check - will tackle other items too)

image

francislavoie:
I think it's fine, but just reduce the padding between the rows so the race and flag are closer to the name, then can have a larger gap to the next line (division/winner). Make the VS be vertically centered in between the name and race/flag instead of top-aligned directly between the names. Basically it should be a 3-column thing, VS in the middle column vertically centered, and the left/right columns have two rows of content.

The division names need to be reintroduced, it should be possible to fit them with smaller font size

MarkWind85:
Regarding the full length name for leagues.. Well you'd think so.. But it doesnt. So id have to place the fist icon elsewhere if you really want the written out league names.. or go an even smaller font, which would be too much i think.


.ovo-league {
  display: flex;
  align-items: center;
  gap: 6px;

  @media (max-width: 850px) {
    font-size: 0.75em;  /* This applies to the league icon, name, division, rank, etc. */
    gap: 4px;
  }
}

image

MarkWind85:
I made one more tweak; removed padding around the content;  which fixes the line break for league on all but the smaller/more narrow phones (like Iphone SE) 
This should be good i think - move onto the other fixes
image

francislavoie:
Should keep top padding for the layout and side padding for the Season and Replay button row of content, but yeah that's better.

I think the hero level number is proportionally too big in mobile compared to desktop, that could be adjusted too.

MarkWind85:
commit state:
Screenshot_2026-05-28_04-15-13
Screenshot_2026-05-28_04-43-22

francislavoie:
Working on a cleanup pass, here's where I'm at:

image

I decided to get rid of the fist, it wasn't actually making sense, ultimately the point of it was to show the winner but we were already doing that with green/red text in the previous design and that works better here anyway. Allows us to tuck things down and it looks more consistent.

I also decided that the league icons being towards the middle makes sense actually, since we do have the full hierarchy preserved in text at least (if we didn't put "Adept" text first, then having the icon after would be weird if it read like `1 #5` or something which is kinda nonsense without it specifying that it's _Adept 1_).

Mobile looks like this:

image


Also made some adjustments to reflow H2H when small to make it all fit better

image

francislavoie (approved):

PR #1073, opened 2026-05-29
Random race filter

MichaelDeitz:
Added switch for random.

### DIFF
--- a/src/components/player/tabs/PlayerMatchesTab.vue
+++ b/src/components/player/tabs/PlayerMatchesTab.vue
@@ -311,6 +311,17 @@ export default defineComponent({
       },
     ]);
 
+
+    // Rehydrate FIRST, before any reactive setup or store calls
+    const savedPlayerRandom = localStorage.getItem("playerIncludeRandom");
+    const savedOpponentRandom = localStorage.getItem("opponentIncludeRandom");
+    if (savedPlayerRandom !== null) {
+      playerStore.SET_PLAYER_INCLUDE_RANDOM(savedPlayerRandom === "true");
+    }
+    if (savedOpponentRandom !== null) {
+      playerStore.SET_OPPONENT_INCLUDE_RANDOM(savedOpponentRandom === "true");
+    }
+
     const playerIncludeRandom = computed({
       get: () => !!playerStore.playerIncludeRandom,
       set: (val: boolean) => playerStore.SET_PLAYER_INCLUDE_RANDOM(val)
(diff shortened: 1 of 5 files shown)


MichaelDeitz:
@francislavoie

src/components/player/tabs/PlayerMatchesTab.vue
```diff
@@ -251,6 +311,35 @@ export default defineComponent({
...
+
+    // Rehydrate FIRST, before any reactive setup or store calls
+    const savedPlayerRandom = localStorage.getItem("playerIncludeRandom");
+    const savedOpponentRandom = localStorage.getItem("opponentIncludeRandom");
+    if (savedPlayerRandom !== null) {
+      playerStore.SET_PLAYER_INCLUDE_RANDOM(savedPlayerRandom === "true");
+    }
+    if (savedOpponentRandom !== null) {
+      playerStore.SET_OPPONENT_INCLUDE_RANDOM(savedOpponentRandom === "true");
+    }
```
francislavoie: I don't see any localStorage stuff for the race choice, so I don't think it makes sense to use localStorage for this. If we did use `localStorage` already, it would be integrated in the `playerStore` itself, not done at the component level.

So just remove all this, it's fine that it clears filters on refresh. What's important is that state is preserved through page navigation, and it should because it's stored in `playerStore` which is not unloaded when switching pages since this is an SPA (single-page app).

src/components/player/tabs/PlayerMatchesTab.vue
francislavoie: Augment this so that when the switch is on and a specific race is selected (not random), add `/` followed by the RaceRandom.png icon (e.g. OrcRandom.png). The second icon should not have `ml-2`.

src/store/player/store.ts
```diff
@@ -368,6 +394,24 @@ export const usePlayerStore = defineStore("player", {
...
+    persistMatchFilterSettings(): void {
+      if (typeof window === "undefined") return;
+      window.localStorage.setItem(
+        MATCH_FILTER_SETTINGS_KEY,
+        JSON.stringify({
+          playerIncludeRandom: this.playerIncludeRandom,
+          opponentIncludeRandom: this.opponentIncludeRandom,
+        }),
+      );
+    },
```
francislavoie: This can go as well

src/store/player/store.ts
```diff
@@ -15,6 +15,22 @@ const playerMatchesCache = new Map<string, PlayerMatchesResponse>();
...
+    if (!raw) return { playerIncludeRandom: false, opponentIncludeRandom: false };
+    const parsed = JSON.parse(raw);
+    return {
+      playerIncludeRandom: Boolean(parsed.playerIncludeRandom),
+      opponentIncludeRandom: Boolean(parsed.opponentIncludeRandom),
+    };
+  } catch {
+    return { playerIncludeRandom: false, opponentIncludeRandom: false };
+  }
+}
```
francislavoie: This can go

francislavoie (changes requested):

francislavoie (approved):
Thanks! Sorry for the delay to review.

PR #1076, opened 2026-05-29
Handle replay download failures in-page

DanyPell:
## Summary
- fetch replay downloads before triggering the browser file save
- keep users on the match page when the replay API returns an error response
- show specific snackbar messages for rate limiting, missing replays, and generic replay failures

## Root cause
The replay button navigated directly to `/api/replays/{matchId}`. When the backend returned JSON errors from replay generation, the browser replaced the match page with the raw API error body.

## Validation
- `npm run type-check`
- `npm run build`
- local browser smoke test against a failing replay response confirmed the page stayed on the match detail view and showed the retry-later snackbar for 429 responses

### DIFF
--- a/src/components/matches/DownloadReplayIcon.vue
+++ b/src/components/matches/DownloadReplayIcon.vue
@@ -17,6 +17,9 @@
     </template>
     <span>{{ tooltip }}</span>
   </v-tooltip>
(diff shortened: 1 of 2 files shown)


francislavoie:
Looks good!

image

francislavoie (approved):

PR #1077, opened 2026-05-29
Add persisted match table options

### LINKED ISSUE #898: Have options for tables
Maybe we could have a standard setup to have an options system (ideally persisted via a cookie) that allows to configure tables views.

So instead of the "Show heroes" button for finished matches, we'd have some sort of dropdown or whatever where you can check options.

Image

Use-cases could include:
1. Toggle country-flag
2. Toggle showing the flo-server (and pings) as a text instead of hover
3. Toggle heroes
4. Toggle showing the player alias

DanyPell:
## Summary
- replace the standalone Show Heroes button with a persisted Table Options menu for match tables
- add toggles for country flags, visible server/ping text, heroes, and player aliases using match-row data already available in the frontend
- apply the same options to the Matches page and player match-history table

## Notes
- defaults preserve the existing table view: country flags on, heroes/server text/aliases off
- player aliases only render when the match row has an invite name that differs from the displayed player name and BattleTag

Closes #898

## Validation
- `npm run type-check`
- `npm run lint`
- `npm run build`
- local browser smoke test for opening the options menu, toggling server text/heroes, and confirming settings persist across reloads


DanyPell:
Closing for now so this can be reviewed locally before opening a PR.

PR #1078, opened 2026-05-30
Improve game table options

DanyPell:
## What changed

- Combined spoiler-free controls and table display preferences into one Settings menu.
- Polished match filters so dropdowns/buttons feel consistent across `/matches` and player match history.
- Kept match status as a simple `Now Playing` / `Finished` toggle without persisting it between reloads.
- Persisted game mode and the table Settings options, while keeping hero filter choice session-only.
- Added hero portraits to the hero selector and kept the Heroes table option always available.
- Removed the duplicate Server & Pings table setting; server/ping details remain available from the host icon tooltip.
- Added a localized Time since table setting, with hover showing the opposite start-time format.

## Notes

The new locale keys are included in `src/locales/data.ts` and `src/locales/en.ts`. Since `data.ts` is generated from the translation sheet during production builds, those rows should be mirrored into the sheet as part of release cleanup.

## Checked

- `npm run lint`
- `npm run dprint`
- `npm run type-check`
- `npm run type-check-vue`
- `npm run build`
- local smoke check on `/matches` and `/player/ElePhanti%232452/matches`


francislavoie:
Looks good to me! I reverted the chevron change, I found it was just making the filters wider and not really adding much in terms of affordances, the button-dropdowns seem pretty well usable without it and it looks much cleaner without it.

image

image

francislavoie (approved):

PR #1084, opened 2026-06-07
Add player warning admin pages

DanyPell:
Adds admin pages for sending player warnings and managing reusable warning templates.

Admins can search a player, pick a template or custom warning, send it, then track sent, acknowledged, and cancelled status.

Templates can be edited per locale with English fallback.

### DIFF
--- a/src/components/admin/warnings/warningLocaleItems.ts
+++ b/src/components/admin/warnings/warningLocaleItems.ts
@@ -1,0 +1,15 @@
+export const warningLocaleItems = [
+  { title: "English", value: "en" },
+  { title: "Deutsch", value: "de" },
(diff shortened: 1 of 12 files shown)


src/components/admin/warnings/warningLocaleItems.ts
```diff
@@ -0,0 +1,15 @@
+export const warningLocaleItems = [
```
marcoabreu: I think you can generalize this since these are all the clientLocales we support, not limited to the warning system

DanyPell: Yep, moved it to shared client locale options in 5c36f1de.

PR #1090, opened 2026-06-10
feat(ranking): apex leaderboard + progression league navigation (P11-4)

marcoabreu:
Adds the public **apex leaderboard** and makes the ladder's league navigation **progression-aware** for progression modes. Client half of P11-4. Additive — the legacy RP ladder render and navigation are unchanged (branched per mode).

## Apex leaderboard view
New `ApexLeaderboardGrid.vue`: the Grand Master + Master cohort sorted by apex points, with a visible **GM cutoff line** between the GM cohort and Master. Reached by selecting the Grand Master / Master tier for a progression mode; data from `GET /api/ladder/apex` into `rankingStore.apexLeaderboard` (own `apexLoading`). Loading / data / empty states all handled.

## Progression league navigation
For a progression mode the league selector lists the **progression tiers** (Grand Master, Master, then leagues with divisions I–IV; grouped menu). Apex tiers route to the apex view; non-apex tiers render a new `ProgressionLadderGrid.vue` from `GET /api/ladder/progression`.
- **Dedicated `ProgressionLadderGrid`** (not `RankingsGrid`): that endpoint intentionally returns rows without the legacy per-player overview, so the grid renders from `playersInfo` + `progression` only — leaving `RankingsGrid` untouched (RP-safe by construction).
- Gateway selector **hidden** for progression modes (gateway-agnostic); race selector shown only for race-split 1v1 (season ≥ `RACE_SPLIT_START_SEASON`).

## Notes
- New `progression-leagues.ts` (pure tier helpers). i18n strings in `en.ts` (the Sheet-generated `data.ts` is untouched).
- New service methods guard `response.ok` (degrade to empty state, never throw).

## Checklist
- [x] Targets `prj/new-ranking-system`
- [x] Task: P11-4 (Workstreams D+E — website)
- [x] No test runner (website) → verified: `type-check-vue` · `lint --max-warnings=0` · `dprint` · `build` all clean
- [x] Visual: pending a screenshot pass (selector / non-apex page / apex cutoff / RP regression, light + dark, 1440/768) — to be attached; endpoints not yet live so verified with mock data
- [x] RP path **behavior-unchanged** (additive branch; `RankingsGrid.vue` not modified)
- [x] Public-repo boundary: whole diff + commit messages scanned — only league/division/points/apexPoints/Grand Master/Master/cutoff vocabulary
- [x] Per-repo doc updated (`docs/ranking.md`)

### DIFF
--- a/src/components/ladder/ApexLeaderboardGrid.vue
+++ b/src/components/ladder/ApexLeaderboardGrid.vue
@@ -20,32 +20,13 @@
               <league-icon :league="entry.league" />
             </td>
             <td>
-              <div
+              <ladder-player-cell
                 v-for="(playerInfo, pIndex) in entry.playersInfo"
                 :key="playerInfo.battleTag"
-                class="d-inline-block rank-icon-container"
+                :playerInfo="playerInfo"
+                class="rank-icon-container"
                 :class="{ 'ml-3': pIndex > 0 }"
-              >
-                <div
-                  class="player-avatar mr-1 pa-0 race-icon"
-                  :title="getAvatarTitle(playerInfo)"
-                  :style="{ 'background-image': `url(${getPlayerIcon(playerInfo)})` }"
-                ></div>
-                <player-rank-info
-                  :player-id="{ name: playerInfo.battleTag.split('#')[0], battleTag: playerInfo.battleTag }"
-                  :alias="playerInfo.playerAkaData?.name ?? ''"
-                />
-                <div
-                  v-if="playerInfo.countryCode || playerInfo.location"
-                  class="d-inline-block ml-1"
-                >
-                  <country-flag-extended
-                    :countryCode="playerInfo.countryCode"
-                    :location="playerInfo.location"
-                    size="small"
-                  />
-                </div>
-              </div>
+              />
             </td>
             <td class="number-text text-end apex-points">
               {{ entry.apexPoints }}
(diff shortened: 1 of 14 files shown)


src/components/ladder/ApexLeaderboardGrid.vue
```diff
@@ -0,0 +1,251 @@
+<template>
```
marcoabreu, line 1: Please evaluate potential for avoiding code duplication across all these different components. It feels like we're repeating a lot

src/components/ladder/ProgressionLadderGrid.vue
```diff
@@ -0,0 +1,154 @@
+<template>
```
marcoabreu, line 1: Same here

src/helpers/progression-leagues.ts
```diff
@@ -0,0 +1,67 @@
+import { LEAGUE_NAMES } from "@/helpers/progression-rank";
+
+// Apex leagues (Grand Master, Master) have no divisions; every other league has 4.
+export const APEX_LEAGUES = [0, 1] as const;
```
marcoabreu: Do we not have the leagues as proper enum somewhere?

Same for the other consts/varialbes and sstuff? It feels like we're inloning a lot.

src/components/ladder/ApexLeaderboardGrid.vue
marcoabreu: Extracted the shared pieces: a `components/ladder/LadderPlayerCell.vue` (avatar + name + profile link + country flag from a `PlayerInfo`) and a `composables/usePlayerAvatar.ts` (the `hasSelectedIcon`/icon/title avatar logic). `ApexLeaderboardGrid` and `ProgressionLadderGrid` now render player cells via `<ladder-player-cell>` — the duplicated markup + avatar functions are gone from both (net −123/+120 across the two grids).

src/components/ladder/ProgressionLadderGrid.vue
marcoabreu: Same — it now uses the shared `LadderPlayerCell` + `usePlayerAvatar`. I deliberately left the existing RP grids (`RankingsGrid`/`CountryRankingsGrid`) out of this pass: they iterate `player.playerIds` (not `playersInfo`), carry Twitch-live + ongoing-match indicators tied to `player`/match state, and use a different icon fallback — so the shared cell isn't a behaviour-preserving drop-in there without risking the RP-critical render path. Happy to do that consolidation as a follow-up (or in this PR) if you'd like it included.

src/helpers/progression-leagues.ts
marcoabreu: Added an `EProgressionLeague` enum (GrandMaster=0 … Grass=8) in `store/types.ts` alongside `EGameMode`/`ERaceEnum`. The raw league numbers are gone: `isApexLeague`/`isApexTier`, the tier-range bounds (`FIRST_NON_APEX_LEAGUE`/`LOWEST_LEAGUE`/`FIRST_DIVISION`), the `DEFAULT_PROGRESSION_TIER` const, and the grid `league` field types now use the enum. `LEAGUE_NAMES` stays as the display-label source, now documented as indexed by the enum.

PR #1100, opened 2026-07-24
Improve match history player search

DanyPell:
Searching "Grubby" in match history gives you six Grubbys and no clue which one you want. Now the dropdown:

- shows each player's profile avatar
- sorts results by highest current-season MMR (shown on the right)
- on hover: bigger avatar, country flag, top 3 modes with W/L + MMR
- closes on selection and no longer re-searches the picked battleTag

No backend changes — uses the existing bulk personal-settings endpoint + game-mode-stats (prefetch capped at 20 players, lazy on hover past that). Every PlayerSearch usage (admin pages, clan invites) gets it for free.


DanyPell:
Closing for now — reworking the approach based on feedback (opponent-scoped search instead of global enrichment).

PR #1104, opened 2026-07-24
Expose mappedForces in the admin map create/edit form

francislavoie:
## What

Adds a validated JSON editor for `mappedForces` to the admin map create/edit dialog.

`mappedForces` is the only field in the matchmaking service's map document with no UI, so the only way to set it today is to hand-craft an API call. It is also the field with **no server-side validation** — the matchmaking service's express-validator rule is spelled `mappedforces` while the contract field is `mappedForces`, so the rule never matches. Client-side validation is the only validation this field gets.

No backend changes are needed: `GET`/`PUT api/maps` already carry the field, and `MatchmakingServiceClient.UpdateMap` already attaches the `x-admin-secret` header the matchmaking service expects.

## Validation

Covers the shape the backend would fail to deserialize, and names the offending path (e.g. `forces[1].slots[0]: "index" must be a number.`):

- top level is an array
- each force has a numeric `team` and an array `slots`
- each slot is an object with a numeric `index`, and a `color` that may be a number, `null`, or absent — `MapForceSlot.Color` is `int?` and no null-ignoring JSON options are configured, so a slot with no colour round-trips to the client as `null`
- `computers`, if present, has numeric `slot` / `color` / `difficulty` / `race`
- `MapForceComputer`'s fields stay strict — they are non-nullable enums on the backend, so `null` is not valid there

Colours and races are checked as plain integers rather than against enum members — the server owns their meaning. Empty text means `[]`. Save is disabled while invalid. A `Format` button re-pretty-prints.

## Line numbers

Errors name a line, not just a JSON path:

- **Parse failures** take line and column from the engine's own message, with its positional tail stripped so the position is not reported twice.
- **Shape failures** carry a path, which a small scanner (`indexPathLines`) resolves to a line by walking the parsed text. That makes the line correct whatever the formatting — minified, tab-indented, CRLF, leading blank lines — rather than only in canonical pretty-printed form.
- A gutter numbers the lines and highlights the offending one, and a **Go to line** button scrolls to it and selects it.

The editor is a small purpose-built control rather than a `v-textarea`. Vuetify gives `prepend-inner` no top alignment inside textareas and masks the top of `.v-field__input` with a gradient, so a gutter placed in its field would be both misaligned and clipped. Sizing the textarea to its content via `:rows` and scrolling the wrapper instead means the gutter cannot desync from the text without any scroll syncing, and `wrap="off"` keeps one line to one row (horizontal overflow stays inside the textarea, so the gutter never slides out of view). Height follows the content up to 15 lines — an empty force list is a 3-line box, not a reserved slab.

## Layout

`EditMap.vue` keeps only the map fields. The editor lives in `MappedForcesEditor.vue` and takes the forces via `v-model`, reporting validity through `update:valid`; parsing and line resolution sit in a colocated `mappedForces.ts` (following the `replays/replayTime.ts` precedent) so they are testable without a DOM.

## Supporting fixes

Four pre-existing bugs that this field's editor either depends on or would otherwise expose:

- **`MapForce.slots` was typed `number[]`.** Stale since the `MapsMappedForces` migration converted stored slots to `{ index }` objects; the backend deserializes into `MapForceSlot[]` (`{ int Index; int? Color; }`), so bare numbers would have failed on the way in. Nothing read `.slots`, so the correction is safe.
- **`AdminMaps.configureMap` / `configureMapFiles` used `Object.assign({}, map)`** — a shallow copy, leaving `mappedForces` and `gameMap` shared by reference with the store's row, so edit-then-cancel mutated it. Now deep-cloned. Previously invisible because no UI touched the nested fields.
- **`mapsManagement/store.ts` `updateMap` swallowed errors into an `alert`**, so `AdminMaps.saveMap`'s own `catch` never fired and the dialog closed and reloaded as though the write had landed. Now propagates. This also fixes the same latent bug for the existing scalar fields.
- **`MapsService` did `throw new Error(parsedJsonBody)`**, rendering as `[object Object]`. A shared helper now extracts a readable message from both the bare-string and `{ errors: [{ msg }] }` shapes, applied in `updateMap`, `createMap` and `createMapFile`.

## Deliberately out of scope

`gameMap` stays out of the form. It is already managed by the Files dialog, which derives it from the uploaded `.w3x`'s parsed header, and its `sha1` / `crc32` / `checksum` / dimensions must match the file bytes. Hand-editing it would be wrong.

Also left alone: `maxTeams` is submitted as a string (`v-text-field` with no `type="number"`; the matchmaking service's `isNumeric()` accepts `"2"` and stores it as a string), and `getAllMaps` builds `api/maps?&filter=...` with a stray leading `&`. Both pre-existing and unrelated to this field — happy to fold either in if you'd prefer.

## Testing

`npm run lint`, `npm run dprint`, `vue-tsc --noEmit` and `npm run build` all clean.

**Against the test environment:** saving works, and clearing the field (emptying it, then saving) works.

**In a browser against the real component:** gutter line spacing 17.906px vs a 17.92px line-height, the first gutter line aligned to the textarea's first text line to the pixel, 74 gutter lines for 74 content lines, box height tracking content (1 line → 54px, 10 lines → 187px, 26 and 162 lines both → 269px, the 15-line cap) and scrolling once capped while the textarea itself never scrolls vertically, so gutter desync is structurally impossible. A 400-character line scrolls horizontally inside the textarea with the gutter staying put. Error states end to end: a parse error reporting `line 6, column 5`, a shape error reporting `line 5`, an error 32 lines down whose **Go to line** scrolled the wrapper and selected exactly `{ "team": "bad", "slots": [] }`, and a return to valid input clearing the error and re-enabling Save. After the split into `MappedForcesEditor`, the `save` payload was confirmed to carry the edited forces alongside the other map fields, with `update:valid` correctly gating the Save button.

**Logic**, exercised through `mappedForces.ts`'s public exports: line indexing on canonical pretty-printed, minified, tab/CRLF/leading-blank and awkwardly formatted documents, including strings containing `{`, `}`, `[`, `]`, escaped quotes and backslashes, plus negative numbers, exponents and empty containers; validation of valid single/multi force input, colours, computers, empty and whitespace input, malformed JSON, non-array top level, bare-number slots, missing and non-numeric `team`, non-array `slots`, missing `index`, bad `color`, non-array and incomplete `computers`, index reporting on the second force, and null handling throughout — null slot colour accepted alone and mixed with real colours, while null `index`, `team`, `slots`, force, top level and null computer `color`/`race` are all rejected. Valid input round-trips unchanged and `null` is preserved.

**Least exercised:** the two error-surfacing fixes only run when the server rejects a save, which the manual pass above did not hit. Flagging that rather than implying full coverage.


francislavoie:
Example:

image

Has some basic validation for syntax or data-type issues:

image

FYI @BogdanW3

BogdanW3:
Looks good to me. Since you mention that there is some (invalid) vslidation on MM side, it would be nice to get that part fixed as well, as a separate effort

BogdanW3 (approved):

PR #1106, opened 2026-07-25
Improve the admin maps UX

francislavoie:
Reworks the Manage Maps screen and its dialogs. Screenshots are from a local run with fixture data, dark theme.

## Manage Maps

image

- **Status is derived, not just the flag.** `api/ladder/active-modes` already returns each active mode with its map pool, so a map reads as **Ladder** (with a tooltip naming the modes), **Custom**, or **Disabled**. The old Enabled/Disabled pair is gone.
- **Enable/disable is a row action**, next to the state it changes rather than beside the actions that open dialogs. It is held for ladder maps, because the matchmaking service refuses to disable a map an active mode uses ("Unable to hide map with ID N because it is currently active in the ladder pool"); the tooltip says so.
- **Filters**: free-text search, a category autocomplete built from the categories in use, a status multi-select (chips coloured like the table), and "only maps without a file".
- Row actions are labelled with tooltips, and the table uses a fixed layout so expanding a row no longer re-measures and shifts every column.
- New maps start disabled — a map with no file is not playable anyway, since the matchmaking service builds its in-game list with `.filter(x => x.gameMap)`.

## Map file details

image

The file list already carries each file's parsed metadata, so no extra requests are needed to show what a file actually contains: author, size, slots, forces, hashes. `num_players` counts every lobby slot, so the breakdown separates the ones a person can take from the computer-controlled ones ("16 total (8 players, 8 computers)"). Warcraft 3 colour markup (`|cAARRGGBB … |r`, `|n`, `||`) is parsed into spans — never `v-html`, since the text comes from map files.

It appears wherever map settings are edited: expandable rows in the maps table and the file list, the edit map dialog, and the select-file confirmation.

## Edit map

image

Name and category on one row, then ID / max teams / disable. Category is a combobox suggesting categories already in use but still accepting new ones; ID and max teams are number inputs; the ID hint is visible instead of hidden in a tooltip.

## Edit map files

image

- The file the map currently uses is highlighted and shows a **Selected** chip instead of a pointless Select button, and the list opens scrolled to it.
- **Download** per file, via the same unauthenticated update-service route the launcher uses (`api/maps/download?mapPath=…`).
- A large drag & drop target replaces the thin file input, with a real progress bar during upload (`createMapFile` moved from `fetch` to `XMLHttpRequest`, since fetch cannot report request-body progress). Uploads stay sequential.
- "Store the file as" is prefilled with the picked file's name and only sent when changed, so the default path still lets the backend use the uploaded name.
- Uploading a name the map already stores is caught before the request: the update service rejects duplicates rather than overwriting.

image

`window.confirm` is replaced by a dialog naming the current and incoming file, with the incoming file's details. Warnings there are informational only — a deliberate layout change is exactly when a file and its configuration disagree — so selecting stays enabled.

## Bulk upload

image

- **Detection runs when files are picked**, before uploading: each file shows its map, category, the file it would replace and a status.
- Files the filename could not resolve are **fixable in place** with a map picker, instead of renaming and starting over.
- Name conflicts — with what the target map already stores, or with another file in the batch — hold that file back with an explanation; the rest of the batch still uploads.
- Per-file "store the file as" override, per-file and overall progress, "Upload & select" to run both steps, and duplicate-map-id warnings.
- The dialog keeps its per-row confirmation after selecting instead of closing on you.

## Notes

- No backend changes. Everything here uses existing endpoints.
- `window.alert` on these screens is replaced by snackbars and inline alerts.
- Checked with `npm run lint`, `npm run dprint`, `vue-tsc --noEmit` and `npm run build`, and driven manually against fixture data in light and dark themes.


BogdanW3:
From the screenshot, the paths when replacing the map file might not both be fullpaths? But otherwise this looks great!

francislavoie:
Replacing (overwrite) isn't allowed. But yes it's just the filename, the server chooses where to store it now, pretty sure.

PR #1107, opened 2026-07-27
fix(admin/lag-reports): surface telemetry failures and refetch on id change

marcoabreu:
## Problem

Action latency on the lag-report detail page is loaded by a **second** request, `GET api/player-match-telemetry/by-game/{gameId}`, independent of `GET api/lag-reports/{id}`. The action-latency traces are gated on a single condition in `LagReportContinuousMonitoring.vue`:

```ts
if (!props.telemetry) return [];
```

So *any* failure of that second request — network error, non-OK status, or the request never being issued — renders exactly like "this game has no action latency": the traces are absent from the chart and the Transport / Action p50 / Action p99 / Disconnects / Crashed columns all show em dashes.

The store did record the reason:

```ts
} catch (e) {
  this.error = e instanceof Error ? e.message : String(e);
  this.telemetry = null;
}
```

…but **no component ever read `error` or `loading`** — only `.telemetry` and `.fetchByGame` were consumed anywhere. A broken request was therefore indistinguishable from genuinely absent data, with zero feedback in the UI. This came out of a support case where one admin saw action-latency traces for a report and another did not; nothing on the page indicated why, and diagnosing it required reading the bundle and the database.

A second, latent bug in the same load path: `AdminLagReportDetail` fetched only from `onMounted`, with no watcher on `props.id`. Vue reuses the component when just the `:id` route param changes, so navigating between two report ids refetched neither the report nor the telemetry and left the previous report rendered under the new id.

And `loadReport` had no `catch` — only a `finally`. A rejection was unhandled *and*, because `onMounted` awaited it, silently skipped the telemetry step entirely.

## Changes

- **`playerMatchTelemetry/store.ts`** — add `attempted`, plus a `skip(reason)` action for callers that cannot build a request at all. This is what makes a genuine 404 distinguishable from never having asked.
- **`notice.ts`** (new) — pure `telemetryNoticeFor()` mapping store state to a message, so the branching is testable without mounting the page.
- **`AdminLagReportDetail.vue`** — render the notice; replace `onMounted` with `watch(() => props.id, …, { immediate: true })`; reset the telemetry store per load.
- **`lagReports/store.ts`** — record `selectedReportError` and clear `selectedReport` up front so a failed or changed load cannot leave stale data on screen.

Resulting states:

| State | Shown |
|---|---|
| Request failed (throw / non-OK) | **error** — `Action-latency telemetry unavailable: <reason>` |
| No request possible (no game id) | **warning** — `…: the lag report did not include a game id.` |
| Fetched, genuine 404 | **info** — `No action-latency telemetry was recorded for this game.` |
| Loaded, or in flight | nothing |

## Test plan

- [x] 5 new unit tests for `telemetryNoticeFor()`, written failing-first
- [x] Full suite green — 21/21 via `npx tsx --test` (5 new + 16 existing)
- [x] `npm run lint` (`--max-warnings=0`), `npm run dprint`, `npm run type-check-vue`, `npm run build` all clean
- [ ] Manual: report **with** telemetry — traces render, no notice shown
- [ ] Manual: report **without** telemetry (404) — info notice, chart still renders client ping
- [ ] Manual: force a non-OK response — error notice shows the status
- [ ] Manual: edit the `:id` in the URL between two reports — both report and telemetry refetch

## Notes

- This repo has `.test.ts` files but no `test` script in `package.json`; they run via `npx tsx --test`. Wiring a `test` script is worth doing separately.
- No behaviour change when telemetry loads successfully — the notice is `null` in that case.


marcoabreu:
Both findings were valid — thanks. Fixed in 83197bdb.

**P1 — stale route loads.** Confirmed, and it was introduced by this PR: with `onMounted` alone `load` ran once, so adding the watcher made concurrent loads reachable for the first time.

Worth noting a component-only guard would not have been sufficient, since `loadReport` assigns `selectedReport` *inside* the store — a superseded response would still commit there. So the guard is applied in both places: each store stamps a request with a sequence number and commits its result, its error, **and** the loading flag only while that request is still the newest. `reset()` and `skip()` bump the sequence as well, so an in-flight response cannot land after either. The component keeps its own token to avoid firing a telemetry request for a report the user has already navigated away from.

**P2 — inspector markers.** Confirmed; `inspectorLeftMs`/`inspectorRightMs` are absolute wall-clock timestamps, now cleared as part of the per-id reset alongside the telemetry store.

Verification: `dprint check`, `eslint --max-warnings=0`, `vue-tsc --noEmit`, `vite build` clean; 21/21 tests pass.

One caveat on coverage: neither store is unit-testable today because both transitively import `API_URL` from `src/main.ts`, so importing them in a test boots the app entry. The sequence guards are covered by review and the type/build gates only. Breaking that import cycle would make these stores testable and is worth doing separately.

marcoabreu:
Root cause found — the surfaced error did its job immediately:

> `'fetch' called on an object that does not implement interface Window`

That is Firefox's wording for the WebIDL receiver brand check. `PlayerMatchTelemetryService` stored the native fetch on the instance and called it as `this.fetchImpl(...)`, which invokes it with the **service** as the receiver. Firefox rejects that; Chromium accepts it. Hence action latency was missing for Firefox admins only, with no request ever reaching the network — and the `TypeError` was swallowed by the store's catch, which is what made it invisible.

Fixed in be3449cd by binding to the global. Injected fetches are left untouched so the seam stays usable from tests.

Added a regression test for the **default, non-injected path**. That gap is the real lesson here: all three existing tests supply a stub fetch, so the production path was never executed by any test and shipped broken. Node does not brand-check, so the test asserts the receiver directly. It fails against the previous implementation (`receiver: the service instance`) and passes now.

I also grepped the rest of `src/` for the same pattern — this was the only occurrence. Every other service calls the module-level `authorizedFetch` helper, a plain function call, which is why only action latency was affected.

Verification: `dprint check`, `eslint --max-warnings=0`, `vue-tsc --noEmit`, `vite build` clean; 22/22 tests pass.

The first two commits remain worthwhile independently of this: without them this bug is invisible, and diagnosing it took a database query plus reading the deployed bundle.

marcoabreu:
Added the shared HTTP layer in 67dcb924, since the root cause was really "every service hand-rolls its own request, so getting it wrong in one place is invisible to the others."

**`AuthorizedClient`** (`src/services/http/AuthorizedClient.ts`) owns URL joining, bearer auth, status checking and JSON parsing, and binds the global fetch in exactly one place. It takes its endpoint plus an optional fetch, which is what makes services built on it testable — `@/config/env` reads `window` at module load, so any service importing `API_URL` directly cannot be imported in a Node test. That is the real reason 8 of 9 service files have no tests.

- `getJson` throws a typed `HttpError` carrying `status`, `method`, `url` and a truncated body
- `getJsonOrNull` maps 404 to null, for endpoints where absent is a legitimate outcome

**Migrated both lag-report services onto it**, which fixed a genuine latent defect in `LagReportService`:

```ts
const response = await authorizedFetch("GET", url, token);
return await response.json();   // a 401 body, typed as LagReportDetail
```

There was no status check at all, so an auth failure or a 500 was parsed and handed to the store as if it were a report — or a page of results. Both now throw, and the list store records the failure so the page reports it rather than rendering an empty table.

**What I deliberately did not do:** migrate the other eleven services off `authorizedFetch`. That helper is not broken (it calls `fetch` bare, so the receiver is correct), and switching them changes error semantics at each of ~108 call sites — e.g. `PermissionService` currently returns `[]` on failure, so throwing instead is a visible behaviour change that needs its own review. It stays, with docs pointing at the client. I chose not to add an `@deprecated` tag yet so those call sites do not all light up before there is a migration plan.

Also wired up `npm test`. The suite previously only ran if you knew to invoke `tsx` by hand, and `tsx` was never a declared dependency — now `tsx@4.23.1` is in devDependencies. Worth adding to CI.

Verification: `dprint check`, `eslint --max-warnings=0`, `vue-tsc --noEmit`, `vite build` clean; **34/34 tests pass** (up from 16 at the start of this PR).

package.json
```diff
@@ -11,6 +11,7 @@
     "preview": "vite build && vite preview",
     "build:prod": "npm run generate-locales && vite build",
     "generate-locales": "node scripts/generate-locales.ts",
+    "test": "tsx --test \"src/**/*.test.ts\"",
```
francislavoie: should probably go with `vitest` instead but 🤷‍♂️ `tsx` is basically a legacy thing

marcoabreu: Agreed, done in c25226a9 — switched to `vitest@4.1.10` and dropped `tsx`.

You are right that it is the better fit beyond just being current: Vitest reuses the Vite resolver and config format, so the `@` alias works without a second toolchain, and it gives us a `jsdom` environment. That last part matters more than it looks here — `@/config/env` reads `window` at module load, which is exactly what currently blocks unit testing any Pinia store or component. With Vitest that is one config line away rather than a refactor, so the store sequence guards in this PR become testable as a follow-up.

Config is in its own `vitest.config.ts` rather than in `vite.config.ts`, so the production Docker build never has to resolve the test runner.

Also wired the suite into `azure-pipelines.yml` as a `vitest` step after the type check, and verified the step can genuinely fail the build rather than being decorative: a deliberately failing test exits 1, the green suite exits 0.

34 tests across 5 files, all passing.

PR #1110, opened 2026-07-27
Add the admin Jobs page

francislavoie:
Front end for the backend's job runner (w3champions/website-backend#541): lists registered jobs with status, progress and last run, and offers Run / Resume / Cancel behind a new `Jobs` permission.

Independent of the other branches in flight — it can merge whenever. It will show an empty list until the backend side is deployed.

### Cards rather than a data table

Other admin pages use `v-data-table`, but a row here carries a multi-line description, a progress bar and an error message, none of which sit well in table cells — for what will only ever be a handful of rows.

### The button says what will happen

"Resume" when a checkpoint exists, "Run again" when the job completed and needs `force`, "Run" otherwise. The runner puts real thought into resume semantics and a generic Run button would waste it. "Start over" is tucked into an overflow menu and **always** confirmed, even for jobs that don't set `requiresConfirmation` — discarding a resume point can throw away hours of work, which is a different kind of destructive from just running the job.

Polls every 2s while anything runs and every 15s otherwise. That is UI refresh only; the runner needs no polling of its own.

### Built on `AuthorizedClient`

Uses the client added in #1107, per its docstring's request that new code prefer it. Worth one note: an earlier revision of this page had `getJobs` swallow failures into an empty array, which renders as *"No jobs are registered"* — a very different claim from *"you don't have permission to see them"*, and exactly the state a user without the permission would hit. It now reports which happened.

### Test plan

- 9 unit tests via vitest (also new in #1107) covering key encoding, the force/reset query and the 409/403/other status mapping.
- `vue-tsc`, `eslint --max-warnings=0`, `dprint check` and a production build all clean.
- Rendered and visually checked in both a light and a dark theme by mounting the component against a stubbed API in every job status. That caught two faults: action buttons wrapping left under long descriptions, and `Interrupted`/`Cancelled` rendering as red errors when neither is a failure.

### Not verified

**The page has never talked to the real API.** Field names rest on the DTO rather than an observed response, and the 403/409 paths are covered by unit tests but not exercised in a browser. Someone will also need to grant themselves the `Jobs` permission in Manage Permissions before the nav entry appears at all.


MarkWind85:
Works smooth, backfill ran properly after registering the job to the backend! :) Though with the local Mongo DB test, the run was so quick that auto-ui-update  never had a chance to show (not realtime but cadence) ; refresh button performed proper though.

One functional issue with the API return (Int) and front end expectation(string). So the final status wasnt correct. 
This is Fixed to string contract in Backend PR: https://github.com/w3champions/website-backend/pull/550

Screenshot with the backend fix:
image

PR #1114, opened 2026-07-31
Stop team order and placement revealing the winner under hide-winner

MarkWind85:
With hide-winner on, the winner stayed readable through layout: the backend
lists the winning team first and the grid rendered that order directly —
winner left, loser right, in every finished match (checked 200 across ten
modes). Survival Chaos also printed the real `#1`–`#4` and listed teams in
placement order.

- With the toggle on, the alphabetically first team takes the left slot and
  the other the right (in team modes a team sorts by its alphabetically-
  first member): left/right is decided by names alone and carries no result
  information. Toggling off restores winner-left exactly.
- Survival Chaos placement renders as `#?` — substituted rather than
  blurred, so the real rank never reaches the DOM.
- `PlayerMatchInfo` captured player props in one-shot refs, so a reordered
  row kept showing the old player's name with the new player's colours.
  They're computeds now. Nothing wrote to or watched these refs, so the
  conversion changes nothing else.


francislavoie (approved):
Yeah, makes sense to me! Thanks

PR #1120, opened 2026-08-18
feat(admin/lag-reports): rework the report list into a triage surface

MarkWind85:
The lag-report list becomes a page you can triage with: a chip filter bar, a grouped view by server node and day, filter suggestions with live counts, repeat-reporter marking, and a summary card when you filter to one player. Fourteen commits, each standing alone. This replaces the existing page wholesale.

**Deploy order: after website-backend#558.** The list always sends a date window, and date filters return HTTP 500 against the current backend (shipped broken in website-backend#506). Everything else it sends is either a new parameter the old backend ignores or an existing one it already understands.

## What an admin will notice

All on a page behind the `Proxies` permission.

| Before | After |
|---|---|
| Unbounded list across the 90-day retention | Default window of today and yesterday (UTC). Presets and manual dates widen it; the grouped view walks seven days further back per step. Clearing filters restores the default window |
| Last visit's filters re-applied silently on arrival | Filters restore silently only when returning from a report detail. Any other arrival shows a prompt row offering them. Deep links win and never prompt |
| "Explicit" | "Submitted". The API keeps its own vocabulary |
| Unknown category in a shared URL dropped locally | The backend's 400 is shown, so a category added after this build still works |

## Where state lives

| Home | Holds | Survives |
|---|---|---|
| URL query | Chosen filters, page | Sharing, bookmarks, detail round-trips |
| `sessionStorage` (`admin-lag-reports-ui-state`) | Filter snapshot, page size, flat/grouped | Reloads and detail visits in this tab |
| `localStorage` (`w3c-admin-lag-reports-prefs`) | Starred filters, visible columns | Everything, per user |

The default window is recomputed on every landing, so a bookmark of the default view stays current and a tab left open across UTC midnight refreshes into the current two days.

## Verify on test

With website-backend#558 deployed, open Admin → Lag reports.

1. Landing shows the two-day window as a chip. Clearing all filters returns to it.
2. Add a filter, reload: a prompt row offers it. Open a report detail and go back: it restores silently.
3. Switch to grouped view by node and by day; bucket counts match the flat list for the same window.
4. Filter to one battleTag: the player summary card appears.
5. Open the page with `?issueCategory=Nope` in the URL: the backend's 400 is shown.
6. Star a filter and hide a column, reload: both persist.

## Reading this

Seventeen files, one idea. Read `src/store/admin/lagReports/filters.ts` first: every filter is a descriptor holding its label, chip text, lifecycle and URL/request serialization, collected in one registry that the bar, the URL sync and the storage sync loop over. After that the filter bar and its six editors read quickly. `AdminLagReports.vue` is the largest diff and is mostly the old inline filter grid coming out.

## Rollback

Revert the squash commit. The two storage keys are namespaced and ignored by the old page.

## Scope

List page only. A follow-up PR improves the report detail page against the same backend deploy and can merge in either order relative to this one.


MarkWind85:
image

https://github.com/user-attachments/assets/05289a50-ce5b-4b6d-b0f7-94206d15f53a

